PR #140, opened 2015-01-06
Ukrainian translation

andriiryzhkov:
Initial commit


afragen:
@andriiryzhkov I missed a string. You should see it in the new .pot file in `develop` branch.

PR #141, opened 2015-01-09
Translate pt-br

valeriosouza:


afragen:
@valeriosouza I missed a string. You should see it in the new .pot file in `develop` branch.

PR #144, opened 2015-01-14
Add French translation

daniel-menard:


afragen:
@daniel-menard I missed a string. You should see it in the new .pot file in `develop` branch.

PR #147, opened 2015-01-30
i18n : plugin description should be translatable

daniel-menard:
Plugin description ("A plugin to automatically update...") should be translatable (see [this](http://stackoverflow.com/questions/25739544/translate-wordpress-plugin-header)).

image

I added a dummy description in a `if (false) {}` block (having it in the code is enough, no need to execute anything while plugin is loading).

I updated the .pot file and the french translation.

PS: I hope my PR is clean, this is the first time I submit 2 pr in a short time, so I created a branch on my fork, but I'm not sure I did it right. Please let me know if there is any problem (I can redo my work once the previous PR is merged if this is simpler for you).

Daniel


grappler:
This is not needed. I have updated the POT with the latest strings and description.
https://github.com/afragen/github-updater/pull/150

afragen:
@grappler was able to add the description as a translatable string. Can you update the PR?

PR #150, opened 2015-01-30
Update POT file

grappler:


afragen:
I just renamed the directory/file structure to move towards PST 4 and I think the POT may need to be re-created. I don't see how to do this with POEdit without loosing the translatable description in this PR

grappler:
I used makepot to create the pot. I would recommend looking into grunt-wp-i18n for the easiest solution. https://www.npmjs.com/package/grunt-wp-i18n

afragen:
Great, one more thing to try and learn 'grunt' :frowning:

PR #145, opened 2015-01-30
Add french translation for string 'no changelog available'

daniel-menard:
Hi Andy!

Here is the french translation for string 'no changelog available'.

(https://github.com/afragen/github-updater/pull/144#issuecomment-72155743)

PR #152, opened 2015-02-04
Add french translation for plugin description

daniel-menard:
Hi !

New PR, replace afragen/github-updater#145 (my working copy was dirty, so I started over...)

The new POT file contains 3 new strings (plugin URI, plugin description and plugin author). 
I'm not sure URI and author are useful in the pot file (they are not meant to be translated, are they?), but it's probably makepot standard behavior.

So, here is the new fr_FR po and mo files, in sync with current pot.

Cheers,

Daniel


afragen:
Thanks Daniel!

PR #156, opened 2015-02-10
Add an action to capture the update infomtion screen and display the readme

toolstack:
When manually updating plugins, the "View version X details." link in the plugin update list breaks as it cannot find the plugin in the wordpress.org repository.

This change "hijacks" the retrieval of the update details for plugins that are hosted on github and instead displays the README.md file.

There are a few things that could be worked on:
    - Only README.md is supported, not readme.md and only from the master branch.
    - It might make more sense to use CHANGES.md (or variations there of).
    - No caching is done at the moment.

### DIFF
--- a/classes/GitHub_Updater/Plugin.php
+++ b/classes/GitHub_Updater/Plugin.php
@@ -67,6 +67,8 @@ class Plugin extends Base {
 		add_filter( 'upgrader_source_selection', array( $this, 'upgrader_source_selection' ), 10, 3 );
 		add_filter( 'http_request_args', array( $this, 'no_ssl_http_request_args' ), 10, 2 );
 
+		add_action('install_plugins_pre_plugin-information', array( $this, 'update_changelog' ) );
+		
 		Settings::$ghu_plugins = $this->config;
 	}
 
@@ -158,6 +160,44 @@ class Plugin extends Base {
 		}
 
 		return $transient;
+	}
+	
+	public function update_changelog() {
+		$dir = WP_PLUGIN_DIR . '\\';
+		$slug = $_GET['plugin'];
+
+		$plugins = get_plugins('\\' . $slug);
+		
+		foreach( $plugins as $val ) {
+			$plugin = $val;
+			break;
+		}
+
+		if( array_key_exists('GitHub Plugin URI', $plugin) ) {
+			if( $plugin['GitHub Plugin URI'] != '' ) {
+				$purl = parse_url( $plugin['GitHub Plugin URI'] );
+
+				$purl['scheme'] = 'https://';
+				$purl['host'] = 'raw.githubusercontent.com';
+				$purl['path'] .= '/master/README.md';
+			
+				$readme = implode( '', $purl );
+
+				$contents = file_get_contents( $readme );
+				
+				//include_once( 'Parsedown.php' );
+				
+				$Parsedown = new Parsedown;
+				
+				include(ABSPATH . 'wp-admin/admin-header.php');
+				
+				echo $Parsedown->text( $contents );
+				
+				include(ABSPATH . 'wp-admin/admin-footer.php');
+
+				exit();
+			}
+		}
 	}
 
 }


afragen:
If you have a `CHANGES.md` or `CHANGELOG.md` in your GitHub repo it will display this information. This is case insensitive.

At this moment I don't parse the readme.txt file but if we were to go that route we could certainly use that if another changelog isn't found.

You should be seeing some sort of information in the view updates screen.

toolstack:
Nope, I don't see anything see attached screenshot.
github updater

afragen:
Can you share the repo so I can test locally?

toolstack:
https://github.com/toolstack/Glot-O-Matic

afragen:
Here's what I did using the latest develop branch.

I installed Glot-O-Matic using the remote install, I then went and changed the local version number downward. An update shows and the following.

screenshot_01

toolstack:
I'll assume it's something on my local dev box then, I still don't see anything.

afragen:
Please let me know if you have similar issues on a live site.

JoeHana:
I actually encounter the exact same issue at the moment. Were you been able to fix it @toolstack ?

toolstack:
I haven't been able to resolve it on my local site and haven't had time to investigate more yet.

afragen:
If you close the 'View details' window and reopen does this persist. Unfortunately I can't reproduce this on any consistent basis. Occasionally I've seen this but closing, refreshing and reopening always seem to fix, or some combination of that.

toolstack:
Mine is consistent and I've found the issue, though this may be specific to my plugin.

Line 106 of src\GitHub_Updater\plugin.php is:

`if ( $response->slug === $plugin->repo ) {`

However in my case the slug is lower case while the repo name is mixed (glot-o-matic in comparison to Glot-O-Matic), which fails the match.

I'm not sure the correct fix here, change the slug to be mixed?  Add a strtolower() around the repo name?

My first thought is to add strtolower() as slugs are lower case but repo names don't have to be.  

Thoughts?

afragen:
@toolstack how can you explain the above image where on my local env th screen displays?

afragen:
Can you tell me what the directory name is? Does it have the extra stuff GitHub adds to zipballs?

toolstack:
The difference is that I've manually installed the plugin, creating the directory when I was developing it as all lower case, where as when you installed it through github updater it would have matched the case to what was in github.

The problem of course is on some systems (like Windows) the file system preserves case sensitivity but doesn't make it unique, so "glot-o-matic" and "Glot-O-Matic" are the same directory.

afragen:
If you change your directory name to match your GitHub repo name does it work? If so then it seems that's the issue. Please let me know. As I'll have to make a notice in the README.

toolstack:
Yes, it works if I rename the directory as well.  However this might be a pretty common problem that is better to be fixed in code.

Slugs are always lower case where as github repo's aren't.  It might be better to do a strcasecmp() instead of ===.

JoeHana:
I'm with toolstack at this point. Repo Names can vary so this issue should be considered by this script, rather than forcing to rename all repos to meet the necessary criterias.

I found that the update screen displayed on one site while it didn't on the other. Both sites are hosted on different servers of different companies. Might be anywhere a specific requirement which needs to be met but which isn't commented?

afragen:
If and when an update is made GitHub Updater will rename the directory correctly. Try it by decreasing the local version number. You might end up needing to reactivate the plugin when this happens.

As this happens it would seem that the issue is more of an initial install of the plugin and this can be solved by having the user install the plugin via GitHub Updater too. This will rename the directory correctly. As some environments may be case sensitive is seems to make more sense to **not** try and fix this in code as there could potentially be a collision.

Thoughts?

toolstack:
You can't count on the filesystem being case sensitive, slugs should always be lower case and compared insensitively to ensure they are unique in the namespace.  

WordPress.org would never allow two plugins in the directory with slugs that only differed by case.

I'd suggest preserving the case of the github repo name is probably the wrong thing to do, instead it should be converted to lower case to match the slug.  When I created my first plugin with WordPress I used mixed case for the directory name, when I published it to WordPress.org the slug was created in lower case and WordPress's upgrade script had no problem matching the mixed case dir name with the lower case slug.

Since many GitHub hosted plugins will have been manually setup in the first place, this could be a pretty common issue.

The other option would be to add another tag to the readme to allow for a different slug than the repo name, which might be better as you could have collisions on GitHub repo names as well (think of a fork of a plugin, the repo name is the same but from different users, you won't be able to have both installed at the same time as they will use the same slug and the updater will place them both in the same directory).

toolstack:
GitHub is case insensitive as well (though the do preserve case), you can't create two repo's that only differ in case in the same account.

Would it perhaps make more sense to use strtolower(<github-userid>-<repo>) as the slug/dir name to ensure collisions don't happen?

afragen:
> I'd suggest preserving the case of the github repo name is probably the wrong thing to do, instead it should be converted to lower case to match the slug. When I created my first plugin with WordPress I used mixed case for the directory name, when I published it to WordPress.org the slug was created in lower case and WordPress's upgrade script had no problem matching the mixed case dir name with the lower case slug.

Can you provide an example in .org where the downloaded plugin directory is mixed case?

> Since many GitHub hosted plugins will have been manually setup in the first place, this could be a pretty common issue.

This can be obviated by using the remote install feature in this plugin. Plus, after any update the directory name will be created **exactly** as it exists on GitHub.

> The other option would be to add another tag to the readme to allow for a different slug than the repo name, which might be better as you could have collisions on GitHub repo names as well (think of a fork of a plugin, the repo name is the same but from different users, you won't be able to have both installed at the same time as they will use the same slug and the updater will place them both in the same directory).

Why would you ever have 2 of the same plugins installed at the same time? I think WordPress innately won't allow this and I agree with that behavior.

It really seems to me that if the directory is named exactly as the repo name then there is no issue. If you can find a scenario where this occurs and there is an issue I will see what can be done to fix it. Right now to fix your particular issue you can rename the directory as it is on GitHub or just decrease the local version, update and it will be done for you.

toolstack:
> Can you provide an example in .org where the downloaded plugin directory is mixed case?

No, WordPress.org doesn't allow mix cased slugs.  When the upgrade happened, it replaced the mixed case dir name with the lower case slug.  On WordPress.org, slugs are ALWAYS lowercase.  But WordPress compares them in a case insensitive manner to ensure matching.

>This can be obviated by using the remote install feature in this plugin. Plus, after any update the directory name will be created exactly as it exists on GitHub.

For new installs, but as many people have done installs in the past by hand it doesn't make much sense to ask them to re-install just to get the upgrade notes working.  Obviously over time as upgrades happen this issue would go away except when users do a manual install.  It would seem unreasonable to assume Updater will be used to do all installs.

>Why would you ever have 2 of the same plugins installed at the same time? I think WordPress innately won't allow this and I agree with that behavior.

WordPress will allow this, though the likely hood is that only one could be active at a time.  Imagine a developer has forked a plugin, they keep both the original plugin installed (but inactive, renaming the directory) to use as reference and monitor changes but install their updated one to develop on.

This could also happen if two users on GitHub create different plugins with the same name.  They are completely separate but share the same repo name and cause a collision.  They may not even have to have the same case to cause the collision depending on the platform being used.  On Windows, where case is preserved by not sensitive, a mixed case repo would overwrite a lowercase repo (or visa versa).

WordPress.org get around this by ensuring the slug is unique, but on github you lose that uniqueness as repo names (used as slugs) only have to be unique within the account.  In fact based on how GitHub works, encouraging forking, repo names are seldom unique across GitHub.

The problem gets worse of course when you include BitBucket, to really get unique slug names for Updater to use you'd have to use [repository]-[account]-[repo].

afragen:
>> Can you provide an example in .org where the downloaded plugin directory is mixed case?

> No, WordPress.org doesn't allow mix cased slugs. When the upgrade happened, it replaced the mixed case dir name with the lower case slug. On WordPress.org, slugs are ALWAYS lowercase. But WordPress compares them in a case insensitive manner to ensure matching.

Even if .org allowed mixed case slugs they would ensure that, case-insensitive, no 2 were the same.

>> This can be obviated by using the remote install feature in this plugin. Plus, after any update the directory name will be created exactly as it exists on GitHub.

> For new installs, but as many people have done installs in the past by hand it doesn't make much sense to ask them to re-install just to get the upgrade notes working. Obviously over time as upgrades happen this issue would go away except when users do a manual install. It would seem unreasonable to assume Updater will be used to do all installs.

Agreed that GHU likely not used to install but as updates occur the issue resolves. This is **only** a potential issue with the initial installation.

>> Why would you ever have 2 of the same plugins installed at the same time? I think WordPress innately won't allow this and I agree with that behavior.

> WordPress will allow this, though the likely hood is that only one could be active at a time. Imagine a developer has forked a plugin, they keep both the original plugin installed (but inactive, renaming the directory) to use as reference and monitor changes but install their updated one to develop on.

I don't think WordPress allows this. I can't even duplicate a plugin directory and change the case. My Mac doesn't allow it.

> This could also happen if two users on GitHub create different plugins with the same name. They are completely separate but share the same repo name and cause a collision. They may not even have to have the same case to cause the collision depending on the platform being used. On Windows, where case is preserved by not sensitive, a mixed case repo would overwrite a lowercase repo (or visa versa).

No collision as a duplicate install couldn't occur.

> WordPress.org get around this by ensuring the slug is unique, but on github you lose that uniqueness as repo names (used as slugs) only have to be unique within the account. In fact based on how GitHub works, encouraging forking, repo names are seldom unique across GitHub.

But the use case for an individual trying to install 2 forks of the same plugin are excessively small. Is this something that you personally do?

> The problem gets worse of course when you include BitBucket, to really get unique slug names for Updater to use you'd have to use [repository]-[account]-[repo].

I'm not really up to fixing theoretical problems. It would seem that the simplest solution and likely the best is to ensure that the directory and repo names are exactly the same. I do understand that you perceive this as an issue, but honestly I've never heard of it in the wild.

toolstack:
I think there are two items we're talking about:

1) Mixed case slug names.  The current code doesn't match insensitively, that should be an easy fix and I don't see any downside to it.

2) Large scale namespace collisions.  This isn't a problem at the moment, but it will eventually become an issue as more plugins are used from different repos.  If you don't want to address it now, that's fine, just keep it in mind for later as it will eventually happen.

afragen:
> 1) Mixed case slug names. The current code doesn't match insensitively, that should be an easy fix and I don't see any downside to it.

Agreed. Released in v4.1.3

> 2) Large scale namespace collisions. This isn't a problem at the moment, but it will eventually become an issue as more plugins are used from different repos. If you don't want to address it now, that's fine, just keep it in mind for later as it will eventually happen.

I think that, while possible, this is a theoretic problem as it's unlikely that a user will be able to install 2 plugins that use the same directory name. If their file system does permit this then the above "fix" **will** cause issues and should be reverted.

PR #169, opened 2015-02-26
add swedish language files

Adevade:
I noticed there are some strings missing from the .pot-file, for example ´Bitbucket Username´ and ´Password´.


afragen:
There may actually be a few more strings in the POT file. I just updated it last evening. Please check and thanks very much!

PR #170, opened 2015-02-26
update and add swedish translation

Adevade:
New files from the latest POT-file. 

However, the translation strings are still missing for `Bitbucket Username` and `Bitbucket Password`


afragen:
You are correct. My fault. I'll fix and upload.

PR #176, opened 2015-03-07
added filters to base -> init function to allow other plugins or themes ...

scofennell:
...to define which users can ccess the plugin settings

### DIFF
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -83,13 +83,34 @@ class Base {
 	 * for proper user capabilities.
 	 */
 	public static function init() {
-		if ( current_user_can( 'update_plugins' ) ) {
+
+		/**
+		 * The prefix for filters within this function.
+		 * @afragen, you might have a better way of doing this elsewhere in the plugin,
+		 * this is just how I like to do it.
+		 */
+		$filter_prefix = sanitize_html_class( __CLASS__ . '_' . __FUNCTION__ . '_' );
+
+		// Determine if the user can update plugins, and pass the verdict through a filter.
+		$plugins        = current_user_can( 'update_plugins' );
+		$plugins_filter = $filter_prefix . 'plugins';
+		$plugins        = apply_filters( $plugins_filter, $plugins );
+
+		if ( $plugins ) {
 			new Plugin;
 		}
-		if ( current_user_can( 'update_themes' ) ) {
+
+		// Determine if the user can update themes, and pass the verdict through a filter.
+		$themes        = current_user_can( 'update_themes' );
+		$themes_filter = $filter_prefix . 'themes';
+		$themes        = apply_filters( $themes_filter, $themes );
+
+		if ( $themes ) {
 			new Theme;
 		}
-		if ( is_admin() && ( current_user_can( 'update_plugins' ) || current_user_can( 'update_themes' ) ) ) {
+		
+		// If the user has either plugins or themes, give him settings as well.
+		if ( is_admin() && ( $plugins || $themes ) ) {
 			new Settings;
 		}
 	}


afragen:
@scofennell I do want this to work with WP Engine and everywhere else. I think that by adding a filter we make it more difficult. I'd rather just incorporate a solution.

Would a better solution for WP Engine be something like

```php
private function wpengine_can_update() {
    return ( false !== stristr( $_SERVER['HTTP_HOST'], 'wpengine' ) ) && current_user_can( 'manage_network' );
}
```

and then add `$this->wpengine_can_update()` to `init()`

afragen:
What happens if WP Engine user is on single site install?

scofennell:
Hey Andy,

My apologies, I started this dialogue with you under the false assumption that this was going to be a common problem for Wp-Engine people.  In reality, the problem in our environment is something that could happen anywhere, and probably will not happen for most Wp-Engine people.  It occurs because we have this in our wp-config:

define( 'DISALLOW_FILE_MODS', TRUE );

I'm happy to hear that you are interested in a solution that is simpler than having to do a filter.  I think the best bet is to write a function that determines the capability needed to use your plugin, depending upon single VS multi.  I do something like this in our plugins:

function lxb_mct_get_settigs_cap() {
    if( is_multisite() ) {
        return 'manage_network';
    } else {
        return 'manage_options';
    }
}

Then you can use the result of that function when you call current_user_can elsewhere in your plugin.

I see that you are trying to offer really granular permissions, where some users might be able to touch themes and others might be able to touch plugins.  Personally, I don't know how useful that is.  And the side effect is that you are biting people like us.

afragen:
With `define( 'DISALLOW_FILE_MODS', TRUE );` in `wp-config.php` there's nothing I can do.

> Disable Plugin and Theme Update and Installation
> This will block users being able to use the plugin and theme installation/update functionality from the WordPress admin area. Setting this constant also disables the Plugin and Theme editor (i.e. you don't need to set DISALLOW_FILE_MODS and DISALLOW_FILE_EDIT, as on its own DISALLOW_FILE_MODS will have the same effect).

> define( 'DISALLOW_FILE_MODS', true );

scofennell:
Yep, exactly, that's why I made this pull request -- so that we could use your plugin even with that constant in our wp-config.  To be clear, in order to make my PR work, I am also running the following snippet in my functions.php:

```php
function sjf_ghu_themes( $in ) { return TRUE; }
add_action( 'FragenGitHub_UpdaterBase_init_themes', 'sjf_ghu_themes' );

function sjf_ghu_plugins( $in ) { return TRUE; }
add_action( 'FragenGitHub_UpdaterBase_init_plugins', 'sjf_ghu_plugins' );
```

It feels a little silly that I'm doing work to remove access only to grant it back, which is why I posed the `manage_options` \ `manage_network` approach as a simple alternative.

If this whole thing seems a little silly, I can and probably should go back to my side and see just how important it is that we run that constant.  You probably won't get this request from any other users ever again.

grappler:
The way I see it github-updater is working as it should. I have added `define( 'DISALLOW_FILE_MODS', TRUE );` to a production site as I do all of the updates in the local site and then push the updates to the production site via git.

WordPress explains this constant as: 
>This will block users being able to use the plugin and theme installation/update functionality from the WordPress admin area.

@scofennell You might want to have a look at `DISALLOW_FILE_EDIT`

afragen:
@scofennell it seems that you're attempting to bypass restrictions that you have put in place to prevent specific users from updating plugin's/themes while still allowing that functionality for yourself.

Perhaps a better solution would be to create a new role for these users that doesn't have the capabilities of `update_plugins` and `update_themes`. Name it something like 'Site Manager' so they have the illusion of control. :wink:

scofennell:
@afragen Yeah We have occasionally said that we need something like "superduperadmins", or just some other role in the hierarchy.  Maybe that is part of the solution here.  I'll go back to my side and see what we can do to avoid being an edge case when it comes to user caps.

@grappler Thank you for that suggestion -- we do currently run `DISALLOW_FILE_EDIT` as well.

PR #177, opened 2015-03-08
Set Base->$options default as an array to avoid PHP warning.

benosman:
When installing the plugin on my dev environment I get the following warning:

```Warning: array_diff_key(): Argument #1 is not an array in plugins/github-updater/src/GitHub_Updater/Settings.php on line 301```

The issue seems to originate on line 77 of Base.php, in the constructor where the get_site_option was not given a default value.

I also changed the declaration on line 63, just to make it clearer that $options should be an array.

### DIFF
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -74,7 +74,7 @@ class Base {
 	 * Loads options to private static variable.
 	 */
 	public function __construct() {
-		self::$options = get_site_option( 'github_updater' );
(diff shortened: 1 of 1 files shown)


afragen:
Good catch, but are both of these changes necessary?

benosman:
The second one ( default for get_site_option) suppresses the error, so only that one is required.

afragen:
Can you update the PR accordingly? Thanks.

benosman:
Sure, now it's updated :)

afragen:
Thanks!

PR #178, opened 2015-03-09
[i18n] - [l10n] > update + French translations v.4.2.0.1

fxbenard:
I let you update the pot. If I do it on my side it will replace your last translator and creator settings.


afragen:
Thanks. I'll get the POT updated.

PR #184, opened 2015-03-12
Changed strings to be compatible with other languages

pedro-mendonca:
In some cases like portuguese the text is slightly different, example:
English: [theme_name] URI
Portuguese URI do [theme_name]

Having the strings as complete sentences makes it possible to translate it in every way.


afragen:
Thanks Pedro!

pedro-mendonca:
I've redone one of those entries to keep it coherent with your code, made a new pull.

Issue #148, opened 2015-01-30
i18n: enhancements suggestions


daniel-menard:
**1. Looking at the code, I found two strings which are *maybe* missing translation**

```php
/includes/GitHub_Updater/Bitbucket_API.php :
- 142: $response->message = 'No tags found';  
- 227: $response['message'] = 'No changelog found'; 
```
They are not printed directly but I guess they are shown somewhere else. Should they be translated?

**2. It's hard to translate pieces of sentences**

It's easier to translate full sentences (see [Otto's Second Law](http://ottopress.com/2012/internationalization-youre-probably-doing-it-wrong/)).

Example:
```php
// includes/GitHub_Updater/Theme.php:306
<?php _e( 'Current version is up to date. Try', 'github-updater' ); ?>
<a href="#" onclick="jQuery('#ghu_versions').toggle();return false;">
    <?php _e( 'another version?', 'github-updater' ); ?>
</a>
```

Should be something like:

```php
printf(
    __( 'Current version is up to date. Try <a %s>another version</a>?', 'github-updater' ),
    'href="#" onclick="jQuery(\'#ghu_versions\').toggle();return false;"'
);
```

Same problem at line `Theme.php:296` (see below).

**3. Avoiding redundancy**

Line at `includes/GitHub_Updater/Theme.php:306` has same problem as above (six strings for 1 sentence):

```php
<strong><br /><?php _e('There is a new version of&nbsp;', 'github-updater' ); echo $theme->name; _e( '&nbsp;available now.', 'github-updater' ); ?> <a href="<?php echo $details_url; ?>" class="thickbox" title="<?php echo $theme->name; ?>"><?php _e( '&nbsp;View version&nbsp;', 'github-updater' ); echo $theme->remote_version; _e( '&nbsp;details', 'github-updater' ); ?></a><?php _e( '&nbsp;or&nbsp;', 'github-updater' ); ?><a href="<?php echo $update_url; ?>"><?php _e( 'update now', 'github-updater' ); ?></a>.</strong>
```

But I think we can re-use the translations used at line 218 (they produce almost the same output):

```php
// new L296
printf(
    __( 'GitHub Updater shows a new version of %1$s available.&nbsp;', 'github-updater' ) . 
    __( '<a %2$s>View version %3$s details</a> or <a href="%4$s">update now</a>.', 'github-updater' ),
    $theme->name, 
	sprintf(
		'href="%1$s" class="thickbox" title="%2$s"',
[...]

PR #185, opened 2015-03-13
[l10n] > French update v.4.2.0.4

fxbenard:
@daniel-menard you can test it out
@afragen well done


daniel-menard:
Hi @fxbenard!

You're too fast for me, nice work!

I can't test now, but I've read [the pot file](https://github.com/fxbenard/github-updater/blob/19d1259a91001b67854fbe7702691e80363ee269/languages/github-updater-fr_FR.po) online. Here are some comments/suggestions, most of them are details so feel free to adapt or ignore:

- L24: 
"MU-plugin", not "my-plugin" 
(I'm probably the culprit of this typo!)

- L55: difficult to read
"La limite sur le nombre d'appels à l'API GitHub sera réinitialisée dans %1$s minutes."

- L60: 
"Il y a probablement une erreur dans les réglages de GitHub Updater"

- L76:
There's no trailing " : " in the original string

- L81: This one is not easy, it's the title of the settings section. I propose:
"GitHub Updater : %s"

- L90:
"Nom de domaine du dépôt distant"

- L109 : +private -"laissez vide" :
"Entrez le jeton d'accès GitHub (GitHub Access Token) à utiliser pour les dépôts privés."

- L184: missing space before the last "%s".
"%sRevenir à la version : %s"

- L219: typo ( "<")
"%sVoir les détails de la version %s %s ou %smettre à jour maintenant%s."

- L241: difficult to read
"à la v.%2$s. En savoir plus sur%3$sles possibilités de mise à jour%4$s."

off-topic:
I didn't know about wp-translations.org, very nice initiative!
(BTW : the link to wp-translations.org is broken in the about page of your personal blog)

Daniel

fxbenard:
@daniel-menard comments integrated - 
Thanks for the review
and thanks for the broken link notification - It's now corrected

(BTW: you're welcome to join WPT)

afragen:
Hey guys, first thanks. Let me know when there's agreement on this PR and I'll merge.

daniel-menard:
You can merge, it's ok for me, fxb really did a good job!

afragen:
Thank you both. I'll try not to mess with the strings for a bit. :wink:

afragen:
Though #186 just fixed some strings in the plugin so some adjustment may be needed. :frowning:

daniel-menard:
I think you should not accept PR against the pot file which is, I presume, directly generated from the source (corrections should go in the code), only translated po/mo.

Would it be possible to rollback #186 and selectively merge it after #185?
Or, better, would it be possible for Pedro to fill a clean PR against "trunk+185-186" ?
(I'm a newbie with git, so I don't really know what is possible or not...)

Some remarks : 
- [this correction should not be merged](https://github.com/pedro-mendonca/github-updater/commit/d124ef7c91ba08230315495cbabd39c70c77862f)
it's not a typo

- [this one](https://github.com/pedro-mendonca/github-updater/commit/b4ad83c98a43231bc86fda1787a70c64a2275267) looks similar to what you did to improve he strings, so I'm not sure it is necessary anymore.

Cheers,

Daniel

pedro-mendonca:
I did correct the source code also, in wich I based the consequent .pot update.

This one https://github.com/pedro-mendonca/github-updater/commit/b4ad83c98a43231bc86fda1787a70c64a2275267 is very similar to another bit of code in the same file, and now they are coherent.

In this https://github.com/pedro-mendonca/github-updater/commit/d124ef7c91ba08230315495cbabd39c70c77862f it should be "Read more information about how you can update." and not "Read more information about (s)how you can update."
In "%3$sshow" %3$s is string and the extra s was probably left when changes from %s to %3$s.

pedro-mendonca:
About the %3$s mentioned above, it's not 'show' but 'how'.
You can spot the origin of the problem here:
https://github.com/afragen/github-updater/commit/317599da95b8a11bfe33605ad598cacac1624967

afragen:
> About the %3$s mentioned above, it's not 'show' but 'how'.
> You can spot the origin of the problem here:
> 317599d

It was a typo

daniel-menard:
> I did correct the source code also

ok

> This one is very similar to another bit of code

I'm sorry, I didn't look close enough, I though Andy already did this correction. 
In fact, this is exactly what I mean in #148 (3. avoid redundancy), thank you for doing the work.

> %3$sshow

Of course, you're right. My bad, I'm sorry for that too.

So, everything would be ok if we update this PR (or fill a new one as it is already merged), based on actual trunk including your corrections?

afragen:
> this one looks similar to what you did to improve he strings, so I'm not sure it is necessary anymore.

I really like readable code and while there aren't specific coding guidelines with respect to this particular format, it is easier to read and identify the particular variables.

General question for translators. Do you prefer `printf` or `sprintf` for strings containing multiple variables?

No, I promise not to go through the code and change everything. :wink:

pedro-mendonca:
Yes, I did those changes to avoid redundancy and to the strings appear as complete sentences to the translator, usually helps a lot to understand the context, and it might make a great difference in some languages.
Pulling out the html tags out of the strings is also a good idea, and safer also.

About printf or sprintf, isn't the difference just 'echo' to show the sprintf? In this case, printf is more sintetic.

afragen:
> So, everything would be ok if we update this PR (or fill a new one as it is already merged), based on actual trunk including your corrections?

It would need to be a new PR as I already merged.

> About printf or sprintf, isn't the difference just 'echo' to show the sprintf? In this case, printf is more sintetic.

The difference is also about being able to put the variables in different order.

pedro-mendonca:
One more thing, I must agree with @daniel-menard that PR to .pot doesn't make much sense as the file should be generated by the author.
I did it at the same time that the typo correction just to save time to the author, trying to minor commit diffs, and so that everyone could update their translations, but sure this is not the ideal workflow. Next time I'll keep .pot updates to myself and wait for the author update.

afragen:
> One more thing, I must agree with @daniel-menard that PR to .pot doesn't make much sense as the file should be generated by the author.
> I did it at the same time that the typo correction just to save time to the author, trying to minor commit diffs, and so that everyone could update their translations, but sure this is not the ideal workflow. Next time I'll keep .pot updates to myself and wait for the author update.

Simplest to just create an issue for the .pot and I'll fix as quick as I can. That said, I greatly appreciate the interest and assistance I receive for this plugin. Thanks all.

pedro-mendonca:
> The difference is also about being able to put the variables in different order.
Ok, I didn't knew this, I just followed the same pattern to make it coherent. So, printf is able to use strings order and sprintf don't, is that it?

> Simplest to just create an issue for the .pot and I'll fix as quick as I can.
Ok :)

afragen:
> > The difference is also about being able to put the variables in different order.

> Ok, I didn't knew this, I just followed the same pattern to make it coherent. So, printf is able to use strings order and sprintf don't, is that it?

I believe so. It's called argument swapping. The syntax for a string replacement is `%s`. To make the string correspond to a specific argument you need to add 'argument number and $' as in first argument `%1$s`

    When using argument swapping, the n$ position specifier must come immediately after the percent sign (%), before any other specifiers
http://php.net/manual/en/function.sprintf.php

daniel-menard:
> It would need to be a new PR as I already merged.

Ok.


> It's called argument swapping

From what I know, both printf and sprintf can use argument swapping:
http://3v4l.org/NBf2v

afragen:
> From what I know, both printf and sprintf can use argument swapping:
> http://3v4l.org/NBf2v

I had no idea as I'm still in the learning stage. But most use I've seen seems to be numbering arguments in `sprintf` and not in `printf`.

As this is the case, it apprears the only difference is **output** formatted string with `printf` and **return** formatted string with `sprintf`

afragen:
The above being the case. Are there variables in the .pot that should be numbered to assist in translations?

pedro-mendonca:
I think the biggest strings are the ones that will capture html code like links are the ones that need it more, because in some languages there are sentences that the order between subject and verb get inverted. For example, if you plan to insert html code in strings in "[Plugin] URI", I would make it swappable, as in portuguese this is inverted to "URI [do plugin]". In doubt, maybe you could add order every time there is more than one ore two.
In case of just "(...) %supdate now%s." is less likely to create a problem, it's a simple open and close of <a> tag in the end of a sentence. Maybe from more than two makes some sense.

daniel-menard:
> printf/sprintf

Yes, the doc says : 
- [printf](http://php.net/printf) : Output a formatted string
- [sprintf](http://php.net/sprintf) : Return a formatted string

Basically, `printf()` is just `echo sprintf()` and they both use the standad library [inherited from the C language](http://en.wikipedia.org/wiki/Printf_format_string).

(there's also [vsprintf](http://php.net/vsprintf) but that's for another day!)


> Are there variables in the .pot that should be numbered to assist in translations?

+1 with Pedro :
It could be a safe precaution for languages where the order of words is different. 

But in your code, it's only worth doing it for strings having more than one variable. 

For example, it's not very useful for the string `GitHub API's rate limit will reset in %1$s minutes.` (Base.php:741) as the sprintf call only have one variable.

But some strings have only one var but are concatenated with others to form the first argument to printf (e.g. Theme.php:232) : 

```php
printf(
    __( 'GitHub Updater shows a new version of %1$s available.', 'github-updater' ) . '&nbsp' . '<a href="%2$s" class="thickbox" title="%3$s">' . __( 'View version %4$s details', 'github-updater' ) . '</a>. <em>' . __( 'Automatic update is unavailable for this theme.', 'github-updater' ) . '</em>',
    $theme_name,
    esc_url( $details_url ),
    esc_attr( $theme_name ),
    $r['new_version']
);
```
Such a line could be rewritten has (just an example) : 

```php
printf(
    __( 'GitHub Updater shows a new version of %s available. ', 'github-updater' ), 
    $theme_name
);
printf(
    '<a href="%1$s" class="thickbox" title="%2$s">%3$s</a>'
    esc_url( $details_url ),
    esc_attr( $theme_name ),
    sprintf(
        __( 'View version %s details. ', 'github-updater' ),
        $r['new_version']
    );
);
printf(
    '<em>%s</em>',
    __( 'Automatic update is unavailable for this theme.', 'github-updater' )
);    
```

More lines, but no more `%x$` in the translation strings, no more concatenation and it looks more legible for my eyes (just a personal taste, I don't like horizontal scroll!)

BTW: in my example, I've included the punctuation in the translation strings (usage of punctuation and spaces before and after symbols vary with the language : "hello!", "bonjour !", "¡hola!") and I replaced the \&nbsp; with a standard space (no need for unbreakable space here, at least in french). 

;-)

afragen:
If this makes translation simpler can you wrap it up in a PR? Thanks

pedro-mendonca:
Leaving a blank space in the end of a string to translate is not a good practice. A lot of translators might not even notice it in programs like Poedit. In the case above, as there are two separate strings, might be better to leave the &nbsp bit.

Of course that if it's made a single string with two sentences, the space will not be missed, but in this case, the second sentence is repeated at least in two places, it's less to translate, no refundancy.

fxbenard:
_x() is then your friend to leave a comment for translators as lot of them will not get the &nbsp;

https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#disambiguation-by-context

or with a description
https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#descriptions

afragen:
I've just updated the .pot, hopefully for last time (for a while). Strings and variables should be much easier.

afragen:
> +1 with Pedro :
> It could be a safe precaution for languages where the order of words is different.

For short strings like `[Plugin] URI` just using `%s URI` is a simple translation to `URI %s`. In this case numbering only helps with multiple variables in the string.

pedro-mendonca:
I think that @daniel-menard just nailed it in https://github.com/afragen/github-updater/pull/185#issuecomment-79224822, with the exception of the space in the end "(...) available.[ ]", that I've explained above already and is alright now. This thread ended with the code very legible and clean. :+1:

daniel-menard:
> Leaving a blank space in the end of a string to translate is not a good practice. A lot of translators might not even notice it

@pedro-mendonca that's a very good point, you're right. 

BTW, what's your opinion regarding non-breaking spaces (U+00A0)? They are important in french typography and I tend to include them litteraly in my code (because my editor show them differently of a normal space), but I imagine they can get unnoticed by translators? Should they be included the old-non utf8 way (\&nbsp;)? or perhaps \&\# 160;?


> This thread ended with the code very legible and clean.

+1, well done, Andy!

pedro-mendonca:
Can you give an example where it's important the &nbsp; ?

daniel-menard:
In France, there is a set of "typography rules", generally edicted by the [Académie française](http://en.wikipedia.org/wiki/Acad%C3%A9mie_fran%C3%A7aise), about that. This is only for France, not for the french language in general (some rules are different in french for Québec for example).

As I'm not able to find an english resource explaining that right now, I will try to give some examples (" " is a normal space, "_" represents a non breaking space):

Punctuation general rules:
  - "single" sign (`.`, `,`) : no space before, normal space after
  - "dual" sign (`;`, `:`, `!`, `?` and so on) : non break space before, normal space after
  - (but there are exceptions...)

(by "non break space", I mean the [usual U+00A0] (http://www.fileformat.info/info/unicode/char/00A0/index.htm) (nbsp) but in theory, we should be using [U+202F](http://www.fileformat.info/info/unicode/char/202f/index.htm) in many cases...)

So one should write "Bonjour, Pedro_! ", not "Bonjour_, Pedro! " or "Bonjour ,_Pedro ! " and so on.

In general, the goal is to avoid bad wrapping (carriage return just before a punctuation sign which takes a space before) :

```
Bonjour, Pedro
!
```

On small screens (smartphones..) it can become important.

Other languages (english, deutch, italian...) generally don't have this wrapping problem because they don't have a space before the punctuation signs.

There are many other usages of nbsp in France, some are:

```
- 37,2_ºC
- 20_€
- dimanche_15_mars_2015
- n°_185
- 1_000 issues
- 2_+_2_=_4
- P._Mendoca
- XVIe_siècle
```

If you're not afraid by french, there are many examples [on this page](https://www.druide.com/enquetes/pour-des-espaces-ins%C3%A9cables-impeccables) and [on this one](http://www.framasoft.net/article2225.html) and some of the current rules are explained on [this french wikipedia page](http://fr.wikipedia.org/wiki/Ponctuation#Les_espaces_autour_des_signes_de_ponctuation_.28r.C3.A8gles_actuelles.29).

There are also rules about the dash ([U+002D](http://www.fileformat.info/info/unicode/char/002D/index.htm) vs [U+2010](http://www.fileformat.info/info/unicode/char/2010/index.htm) vs [U+2011](http://www.fileformat.info/info/unicode/char/2011/index.htm) and so on), but this comment is already getting too long!

daniel-menard:
I finally found some english resources about that:

- http://unicode.org/udhr/n/notes_fra.html
- http://tex.stackexchange.com/questions/64525/comparing-typographic-traditions-for-different-languages
- http://www.mancko.com/typography-punctuation/en/

pedro-mendonca:
I think I understand.
In the end, for languages like portuguese and spanish where the verbs and subjects are inverted than from english, and like this situations you've explained for France, this is why the sentences should always be as complete as possible. Still, leaving a space in the end is still not a good practice. When a string is a complete sentence in english, it's better to keep the ponctuantion within the string and leaving to the translator the choice on how to ponctuate, wether is with nbsp (or alternative) like in France or without it.

PR #187, opened 2015-03-13
Update Italian Language

overclokk:


afragen:
I've just updated the .pot file. Please adjust the PR accordingly. Hopefully the strings will be simpler.

PR #189, opened 2015-03-14
Missing text domain in WPUpdatePhp

daniel-menard:


afragen:
I do realize the test domain is missing here. I've bee working with @coenjacobs on this project and there's no clean way to add a text domain. My challenge is, do we leave it blank or add it when it gets used in a plugin like this one?

pedro-mendonca:
Why won't the text domain be included in every single string?

afragen:
I agree, there's no reason the text domain should be omitted. I'm curious, the string does show up in the .pot file. Does the translation show up if the text domain isn't present? You can test by changing the version number to something like `$updater = new WPUpdatePhp( '6.0');`

daniel-menard:
IIRC, 

Tools like poedit, gettext and so on just look for functions like __(), __e(), __n(), etc (see "X-Poedit-KeywordsList" in the pot file header).

In wordpress, the text domain argument is not mandatory. No text domain means "the wordpress default text domain", so poedit extracts it and this is why it appears in the pot file, but the translations are not loaded because wordpress looks for the string in the wordpress mo files, not in ours.

pedro-mendonca:
The .pot gets the string because the code was browsed by the functions __(), _e(), etc, wether they include or not the textdomain. If the textdomain isn't present it won't get the translation for that particular string.
Still, if you want me to test, where do I use that $updater = new WPUpdatePhp( '6.0'); ?

pedro-mendonca:
:1 @daniel-menard

daniel-menard:
> Still, if you want me to test, where do I use that $updater = new WPUpdatePhp( '6.0'); ?

Just change the required version in `github-updater.php:36` and browse in wp-admin, the string will appear as an admin notice.

pedro-mendonca:
Ok, I've confirmed, after adding the 'github-updater' textdomain, the notice appears translated :)

afragen:
But not translated when no text domain is present?

daniel-menard:
Sorry, I should have said that I've already tested all that (my PR description is... minimal)

No the string won't load without a text-domain.

pedro-mendonca:
Yep, It appears in English.
After textdomain added it gets the translation.
That's what @daniel-menard said, without textdomain it tries to find the string in the main WordPress .mo files. You probably would get lucky with strings like "View" our "Save". Still, it's not a good idea to rely on that. From a translators point of view, getting a 100% translated file it's always what matters, wether the string it's already present in the main .mo files.

afragen:
Thanks for confirming.

PR #190, opened 2015-03-14
New Update Italian Language

overclokk:


afragen:
Thanks

afragen:
I should really look more carefully. PR should only be to PO/mo files. I don't think anything changed but this included many other core files.

PR #192, opened 2015-03-15
String cleanup

pedro-mendonca:
Moved %s unnecessary variable from the end of the string out to the code.

### DIFF
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -747,7 +747,7 @@ class Base {
 						__( '%s was not checked. GitHub Updater Error Code:', 'github-updater' ),
 						'<strong>' . $this->type->name . '</strong>'
 					);
-					echo '&nbsp'.self::$error_code[ $this->type->repo ] ;
+					echo ' ' . self::$error_code[ $this->type->repo ];
 				?>
 				<?php if ( 403 === self::$error_code[ $this->type->repo ] && false !== stristr( $this->type->type, 'github' ) ): ?>
 					<br><?php printf( __( 'GitHub API\'s rate limit will reset in %s minutes.', 'github-updater' ), self::$error_code[ $this->type->repo . '-wait'] ); ?>
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -747,7 +747,7 @@ class Base {
 						__( '%s was not checked. GitHub Updater Error Code:', 'github-updater' ),
 						'<strong>' . $this->type->name . '</strong>'
 					);
-					echo self::$error_code[ $this->type->repo ] ;
+					echo '&nbsp'.self::$error_code[ $this->type->repo ] ;
 				?>
 				<?php if ( 403 === self::$error_code[ $this->type->repo ] && false !== stristr( $this->type->type, 'github' ) ): ?>
 					<br><?php printf( __( 'GitHub API\'s rate limit will reset in %s minutes.', 'github-updater' ), self::$error_code[ $this->type->repo . '-wait'] ); ?>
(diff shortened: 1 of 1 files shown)


afragen:
@pedro-mendonca my habit in this code has been to put `printf( __(…),` all on the same line. Also, given the discussion regarding nbsp should we include it here? @daniel-menard thoughts?

Also, please put spaces before and after concatenation. I think is in WP Code Guidelines.

Thanks.

daniel-menard:
**About nbsp:**
Could be a normal space (not in a translation string, so won't be missed by translators), but it doesn't mind here.

**About coding standards:**
My very personal opinion is that [WordPress PHP Coding Standards](https://make.wordpress.org/core/handbook/coding-standards/php/) are mandatory for code intented to be included in wordpress core (patches and so on).

In my own code, I feel free to not adhere to these which are wordpress specific. In fact, I prefer to keep inline with [PSR-1](http://www.php-fig.org/psr/psr-1/) and [PSR-2](http://www.php-fig.org/psr/psr-2/) which are the de facto standard used by many many frameworks and libraries.

Whatever the standard you use, very long lines is a legibility problem for me (did I say I dislike horizontal scrollbars?) On this point, PSR-2 says: *"There MUST NOT be a hard limit on line length; the soft limit MUST be 120 characters; lines SHOULD be 80 characters or less"*. But AFAIK, WP standards do not impose any limit, so feel free to use what works for you ;-)

pedro-mendonca:
To start, this commit was first about pulling the ending unnecessary %s out of the string.
As I've kept the final ponctuation ":" it's ok for a specific translation to choose to include a space "_" before the ":" in their .po/.mo file.

About the inline vs indented, I've just walk the same coherence path as before, I saw that the author has walked before. As everyone codes a bit differently, whatever pull requests I do I tend to reuse the author pattern.
Here https://github.com/afragen/github-updater/commit/5bedfab6f1eb25b1653d99876d0a0c7d89784267 I've maintained the original inline hard to read code pattern.
After I saw the @afragen commit https://github.com/afragen/github-updater/commit/0ed40c071ef7e3f31abddf8f4e83a47b72d5aa25 and I thought I did wrong in keeping the code inline, as he was writing this indented.
Then, I made this change while taking out of the string on unnecessary %s in this commit https://github.com/afragen/github-updater/commit/b4ad83c98a43231bc86fda1787a70c64a2275267
Here is another example that maked me think that  @afragen was starting to indent where he was rewriting code: https://github.com/afragen/github-updater/commit/0a174b5ed5bede9912eeea28105040916b8f8f2f

This is why I've rewritten this bit of code also as indented, as I rewritten to pull the unnecessary %s out of the string.

Still, I perfectly agree with @daniel-menard, horizontal scroll is a pain, no trackpad here, just a classic scoll wheel mouse :)
Of course I will respect whatever pattern of code the author uses, and as I've said above, I never suggested any pattern that I wasn't already watching in the recent changes to the code.

Now, about the space used I've, I think it will work like this for any language, probably in this case breaking a line is not a problem as it is an error code afrer the ":".
But if for french translation it would be better to have the full '%s was not checked. GitHub Updater Error Code: %s' than I think you should cancel this PR.

I've changed the nbsp to " " (normal space) and corrected the spaces in concatenation in an aditional PR.

pedro-mendonca:
To close this, I'm sorry if I misunderstood the code pattern and you rather have the inline code, it's your call, I will keep writing in whatever pattern you decide.
Please keep this and previous indented coded PR just as suggestions to improve reading the code.
Feel free to rewrite it in wichever pattern you prefer :)

afragen:
I'll update with a few corrections so you can see what I'm referring to. Thanks.

PR #194, opened 2015-03-16
minor tweaks to translatable strings when there is more than one variable

jazzsequence:
Like it says, minor tweak to make use of `%1$s`/`%2$s`/etc when more than one variable is in a string

Ping me if there are more i18n-related issues. :)


afragen:
Thanks Chris

PR #207, opened 2015-03-31
Localization .pot and pt_PT updates

pedro-mendonca:


afragen:
Thanks @pedro-mendonca

PR #218, opened 2015-04-15
added German translations

limenet:
they are valid for de_CH, de_DE, and de_AT

Do you prefer the file(s) to be named `github-updater-de.po` or `github-updater-de_DE.po`?


afragen:
> they are valid for de_CH, de_DE, and de_AT

> Do you prefer the file(s) to be named `github-updater-de.po` or `github-updater-de_DE.po`?

I think I'd prefer using the `de_DE` suffix, however, would this have consequences as it appears the files are valid for several different dialects?

limenet:
I don't know that much about WordPress translations, but since I'm from Switzerland (`de_CH`) I've often noticed translation with `de_DE` not working (of course). WordPress however, lists `de_CH` as one of the official languages when you install it, so I figure instead of forcing Swiss (and Austrian, too, of course) people to use `de_DE` only to make translations work, it'd be nicer if `de` translations just worked.

We could either:

1. Leave it as `-de`
2. "Force" everybody to use `de_DE` by renaming appropriately
3. Triplicating the file for all the dialects

I'd go with option 1. What do you think?

afragen:
If option 1 works correctly for the most number of users and the translations show up it seems more simple.

limenet:
I just tested `de` on a `de_DE` site and it didn't work, unfortunately. So we'll have to either use 2. or 3.

afragen:
Do you want to add it all to this PR or make separate PRs? Probably will need all 3.

limenet:
Sorry for the delay, will add them to the PR, if that's okay with you.

afragen:
@limenet I'm sorry. I just updated the .pot file to include an additional string for an error message. Can you update before I merge?

limenet:
@afragen No problem, I'll do that right away. Since I failed at correctly pulling the changes from your repo, do you want me to close this PR and create a new one to keep it tidy?

afragen:
Whatever works best for you.

afragen:
I see what you mean. Let's close this one and start fresh.

limenet:
see #222

PR #237, opened 2015-05-20
Remove the checkbox if loaded via MU-plugins

pbearne:
add's js to remove the checkbox from the row as well
just to tidy it a bit


afragen:
Great idea thanks. I think I found a simpler method. https://github.com/afragen/github-updater/commit/18fd40b2eeeba83457019ce4bda2c27fed667993

PR #238, opened 2015-05-29
Added the Japanese translation

1shiharat:


afragen:
@1shiharat thanks.

It appears that there are a few strings not translated. Can you fix them before merging?

afragen:
I'm going to merge. We can add the remaining strings later. Thanks!

PR #245, opened 2015-06-07
Add translate for Russia

yumashev:
65% translate for start :)


afragen:
Thanks. 65% is a great start.

PR #246, opened 2015-06-09
Split settings and plugins & themes on wp hook

chefduweb:
On profiling my own plugins with [P3] (https://wordpress.org/plugins/p3-profiler/), I noticed that Github Updater uses a lot of time to complete it's api requests. While most of this needs to gets done in the admin, it seems to get checked on every init-hook in WordPress...

I've split the plugins & themes init-functions with the settings-init, so plugins and themes init on the WordPress hook 'admin_init'. settings still hooks into init, otherwise the settings-page wouldn't fire.

I've added three speed-tests:
one on init (the slowest)
one on admin_init (the fastest, but without the settings-page firing)
and one split up (like in this pull-request) which makes everything work again.

like pull request
with init
with admin_init

### DIFF
--- a/github-updater.php
+++ b/github-updater.php
@@ -68,4 +68,5 @@ new $instantiate;
  * InfiniteWP, ManageWP, MainWP, and iThemes Sync will load and use all
  * of GitHub_Updater's methods, especially renaming.
  */
-add_action( 'init', array( 'Fragen\\GitHub_Updater\\Base', 'init' ) );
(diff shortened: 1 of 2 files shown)


chefduweb:
Updates seem to work as usual, but maybe this needs some more testing.

afragen:
You can't just fire on admin_init as update services like ManageWP, iThemes Sync, etc, won't work.

Only an admin user sees the initial times for the API calls. All API calls are cached and subsequent checks are against the transient.

Issue #242, opened 2015-06-02
Disabling update check on regular requests?


khromov:
When a site has not been visited for a while, GitHub Updater decides to load external data from the GitHub API in the same process as the page load. Dozens of HTTP requests are sent out and the user is just waiting.

Here is what that looks like.
![github-updater-damn](https://cloud.githubusercontent.com/assets/1207507/7948463/6e3d78da-0984-11e5-878d-60edac20264f.png)

25 HTTP requests, 22.67 seconds TTFB (!)

The issues happens when creating the plugin instance and calling get_remote_info():
https://github.com/afragen/github-updater/blob/develop/src/GitHub_Updater/Plugin.php#L64

It would be much better to register a proper WP Cron job that would handle this. In fact, we can see that a WP Cron job also starts when a site hasn't been visited for a while:
![github-updater-damn2](https://cloud.githubusercontent.com/assets/1207507/7948547/11f2c156-0985-11e5-9669-5ae32de99932.png)

This issue is quite severe if you have a lot of GitHub Updater-enabled plugins, I can't currently use it because forcing a random visitors to wait 15-30 seconds for a page load isn't acceptable. Hoping it can be resolved.

afragen:
This only happens when an admin user visits the site and all data is cached for 12 hours. Regular users won't see a performance hit. Test with any of the site speed tests.

I made quick stab at using cron and while I got a cron job set up it wouldn't show updates. I know there must be something more to this but I haven't had the time to figure it out.

**Bottom line:** no regular users will be impacted by this plugin. Please refer to the following code.
https://github.com/afragen/github-updater/blob/develop/src/GitHub_Updater/Base.php#L105-L115

Admin users will only see this once every 12 hours, unless they use the 'Check Again' button.

khromov:
Thanks for the answer. I'll give the WP Cron method a stab and submit a PR, because that's the way WP checks for updates internally, so maybe we can even piggyback on that CRON call.

afragen:
That would be great. If love to see the performance differences. As long as it's functionally identical we could move towards we-cron.

yafred:
@khromov Hi, what is the plugin you're using to analyze performances ?

khromov:
@yafred The plugin in the screenshot is Query Monitor:  
https://wordpress.org/plugins/query-monitor/

@afragen I started doing the Cron refactoring. Not quite ready yet but expect a PR soon. :-)

afragen:
[...]

PR #270, opened 2015-07-29
Update plugins / themes via CRON

khromov:
This PR stems from the discussion in #242 

By introducing the `$force` option to get_plugin_meta() and `$force_meta_update` to the Plugin and Theme class constructors, we no longer make external calls on a regular (admin) page load.

By introducing a number of hooks in the new `background_update()` function, we make sure that the plugin metadata will attempt to refresh on the following conditions:

* Visiting plugin page
* Visiting update page
* When`wp_update_plugins` and `wp_update_themes` hooks run. (These run on the CRON jobs with the same name. WordPress automatically sets up these jobs to run every 12 hours.)

Basically, on the above conditions, the plugin will work as it did before. On all other pages it will *not* attempt to grab data remotely. (That is what the CRON job is for!)

You can try this out by installing the Crontrol plugin and manually running the `wp_update_plugins` and `wp_update_themes` CRON jobs.

I made this as a minimal working proof of concept, using the existing code structure. I haven't tested branch switching but seeing as the changes to the code are minimal, it should work as before. Any feedback welcome.

### DIFF
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -112,6 +112,7 @@ class Base {
 		 */
 		add_action( 'init', array( &$this, 'init' ) );
 		add_action( 'init', array( &$this, 'remote_update' ) );
+		add_action( 'init', array( &$this, 'background_update' ));
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


afragen:
Shouldn't there be a corresponding change to https://github.com/afragen/github-updater/pull/270/files#diff-826f6032956806670f536a25ad599654R90 adding the `$this->force_meta_update` boolean?

afragen:
I'm creating a new branch for this PR so I can test before complete merge. Can you adjust the PR to `wp-cron` branch?

khromov:
GitHub won't let you change branches on a PR but I created a new one here: #272

PR #272, opened 2015-07-29
Update plugins / themes via CRON, Pt. 2

khromov:
This PR stems from the discussion in #242 

By introducing the `$force` option to get_plugin_meta() and `$force_meta_update` to the Plugin and Theme class constructors, we no longer make external calls on a regular (admin) page load.

By introducing a number of hooks in the new `background_update()` function, we make sure that the plugin metadata will attempt to refresh on the following conditions:

* Visiting plugin page
* Visiting update page
* When`wp_update_plugins` and `wp_update_themes` hooks run. (These run on the CRON jobs with the same name. WordPress automatically sets up these jobs to run every 12 hours.)

Basically, on the above conditions, the plugin will work as it did before. On all other pages it will *not* attempt to grab data remotely. (That is what the CRON job is for!)

You can try this out by installing the Crontrol plugin and manually running the `wp_update_plugins` and `wp_update_themes` CRON jobs.

I made this as a minimal working proof of concept, using the existing code structure. I haven't tested branch switching but seeing as the changes to the code are minimal, it should work as before. Any feedback welcome.


khromov:
@afragen We accidentally got the 4f43c2e commit in here due to me merging the develop branch, but it shouldn't matter once you merge the PR into wp-cron and then into develop further down the line. (The commit exists there)

afragen:
OK, i've been correcting a few little things to make but fundamentally I don't see any plugins or themes that are listed for updates and I know I have them.

Does this work? I'm testing locally. It seems as if the `pre_set_site_transient_update_` hooks aren't called.

khromov:
Hey @afragen 

It looks like you have commented out the initialization of the Plugin and Theme classes: https://github.com/afragen/github-updater/blob/wp-cron/src/GitHub_Updater/Base.php#L121

The additional hooks added in `background_update()` only trigger under specific conditions and are not "instead of" the regular initialization, so you still have to keep the old way of loading them inside the `init()` function.

This means that the Plugin and Theme class may be initialized twice. I didn't see a clean way to fix that without some heavy refactoring.

afragen:
Yeah, I did that as I was having duplicate `after_plugin_row` calls. Now fixed.

I guess I didn't understand the `background_update()` in addition to. I thought it was supposed to be instead of.

Can you explain how the background_update works? When and how often should it fire? It seems that background_update doesn't cause the `pre_set_site_transient_` hooks to fire.

I thought the _goal_ was for the `wp_remote_get` calls to be run through wp-cron, sorta like core does with .org plugins/themes. As everything is pretty transiented and the code only runs for admin level users, what improvements are we gaining?

> I didn't see a clean way to fix that without some heavy refactoring, because in the current code you don't assign these classes to a variable so they essentially become impossible to reference. You can fix that by either storing them as a property on the Base class or by implementing a singleton pattern.

I'd like to avoid singleton's at the moment because I really want to learn how to write unit tests for this project and have no idea how, but I heard it's difficult to test singletons, dependency injection was better.

khromov:
@afragen 

The goal of the PR was to stop the `get_remote_info()` from being run on user-facing admin requests. This was implemented by adding a flag into the Plugin & Theme class constructors that made sure the `get_remote_info()` would not run:
https://github.com/afragen/github-updater/pull/272/files#diff-b5d91531d00c16c9a1c8e7a3ba53d0c5R230

But, now we're not updating plugin metadata anymore, so we still have to do that somehow. Queue `background_update()`:  
https://github.com/afragen/github-updater/pull/272/files#diff-b5d91531d00c16c9a1c8e7a3ba53d0c5R136

If we look at the `wp_update_plugins` / `wp_update_themes` hooks first - these two hooks are only run inside a WP CRON request. 

So how does WP CRON work? Whenever an admin user loads WordPress, WP checks whether any CRON jobs need to run (they are scheduled by timestamp), and if they do, WordPress makes a secondary request to itself using the spawn_cron() function. This secondary page load will load /wp-cron.php, which in turn loads another full copy of WP.

Now we're in the secondary CRON request, you can see this as being run asynchronously with the request that spawned it - spawn_cron() uses a cute little trick and sets a very low timeout value when calling the secondary CRON request, the result being both requests run in parallel, see here:
https://github.com/WordPress/WordPress/blob/2f3e567f44e7bc335b6795713ae970b4f37117cd/wp-includes/cron.php#L297

In the secondary CRON request (which is just the wp-cron.php file), WordPress starts to check the internal CRON schedules. There are two schedules (stored in the database) called `wp_update_plugins` and `wp_update_themes` that run every 12 hour. These **are** the default jobs that check for WPorg plugin/theme updates. (You can verify this with the Crontrol plugin.)

**If** any of these two schedules run at this particular WP CRON invocation, they will activate hooks with the same names, and those are the hooks we use in `background_update()` to piggyback on top of the regular update process. That will cause the following code to run:
https://github.com/afragen/github-updater/pull/272/files#diff-b5d91531d00c16c9a1c8e7a3ba53d0c5R149

Since we pass "true" into the constructors, this will trigger `get_remote_info()`. But it will do it in a secondary request, thus not blocking the user from reaching the admin interface. I have verified that clearing all transients and then manually running the wp_update_plugins CRON schedule correctly populates the plugin / theme metadata.

The additional `load-plugins.php` , `load-update.php` and `load-update-core.php` hooks makes sure we refresh metadata when visiting the plugin listing and update pages.

Note that in all these cases, we will in fact initialize the Plugin and Theme class twice. That could potentially cause issues, which is why we probably should only do it once, if not with singletons then by storing them as an object property on the Base class.

Which hooks are not firing properly?

afragen:
I've just updated the `wp-cron` branch to use singletons for Plugin and Theme classes.

I think it's working like it's supposed to but can you test? Thanks.

khromov:
Good work @afragen , we're getting there!

From looking at the code, I don't think you should need row 131-134 here:
https://github.com/afragen/github-updater/blob/wp-cron/src/GitHub_Updater/Base.php#L131

The reason for that is that's already done in background_update(). The problem with your instance function is that it does not allow the setting for $force_meta_update to change. The instance method should look something like this:

```php
	public static function instance( $force_meta_update = false ) {
		$class = __CLASS__;
		if ( false === self::$object ) {
			self::$object = new $class( $force_meta_update );
		}

                //Let us reconfigure the $force_meta_update flag
                self::$object->force_meta_update = $force_meta_update;
		return self::$object;
	}
```

And in __construct of Base: 
```php
		add_action( 'init', array( &$this, 'background_update' ), 11 );
		add_action( 'init', array( &$this, 'init' ) );
```

(This is to make sure background_update runs after init and has a chance of optionally overwriting the $force_meta_update_flag).

See if that works, I can check further this weekend.

afragen:
> From looking at the code, I don't think you should need row 131-134 here:
https://github.com/afragen/github-updater/blob/wp-cron/src/GitHub_Updater/Base.php#L131

I did this as the Settings menu wasn't displaying on other admin pages. I think I'll have to go back to setting a transient for the instances so that they are easily available to `new Settings();`. I should be able to fix that up after work today, on call so not sure when.

The instance method is now `Plugin::instance()` and `Theme::instance()` so the above method will need to be a bit different. I didn't realize we were changing `$force_meta_update` in the instance method? I thought it was a passed value.

Thanks for all you help. I think we're certainly getting closer.

khromov:
If you call `instance()` twice, it will discard the second parameter you pass. Consider the following code:

```php
$foo = Plugin::instance( false );
$bar = Plugin::instance( true);

//Returns "false"
var_dump(Plugin::instance()->force_meta_update);
```

So, we essentially throw away the `$force_meta_update` flag on the second call.

The reason `Plugin::instance()` is called twice is due to hook timing. For example, when you're visiting the Plugins page, `Plugin::instance()` will first be called on the `init` hook, here:
https://github.com/afragen/github-updater/blob/wp-cron/src/GitHub_Updater/Base.php#L114

Then, since we have added the `load-plugins.php' hook in `background_update`, it will be called another time later on when that hook fires. The second time around, now being in `forced_meta_update` we want to change the `$force_meta_update` flag since we now have figured out that - aha, we **should** in fact do remote calls this request.

Maybe having a toggle function would make the functionality clearer.

```php
	function forced_meta_update() {
		Plugin::$object = Plugin::instance()->enable_remote_updates();
		Theme::$object  = Theme::instance()->enable_remote_updates();
	}
```

So, we wouldn't accept any parameter to `instance()` and the default would be no remote updates. The `enable_remote_updates()` would just flip the `$force_meta_update` boolean.

afragen:
While I understand what you're saying, I don't understand the toggle function.

Why is it that if I set 
```php
add_filter( 'init', 'background_update', 0 );
add_filter( 'init', 'init' );
```
`init() ` still fires first. I need `init()` to run after `background_update()`

latest commit to `wp-cron` would work great if the above could happen.

khromov:
It's because `background_update()` doesn't run `forced_meta_update()`, rather, it schedules `forced_meta_update()` to run at a later time using `add_action` and various hooks that we have discussed. 

Init will always fire before those hooks because for example `load-plugins.php` is done at a later time in the bootstrap process, see:
https://github.com/WordPress/WordPress/blob/master/wp-admin/admin.php#L226

I'll have some time in the beginning of next week to look at this further if you haven't cracked it by then! :-)

afragen:
What about adding `init()` to its own cron job with a 60 sec delay or something? Possible?

khromov:
No, hooks can't be delayed in that fashion unfortunately.

afragen:
And that's why I original put in that array of pages to run `instance(true)`. Maybe I do the same thing but then it would only make sense to use `background_update` for `wp_update_plugins` and not on the page loads.

afragen:
I think the latest commit fixes the remaining issues, except I'm now seeing something strange. When doing an AJAX update from the `plugins.php` page the version number seems be replaced with an empty string. This used to work and I'm not sure what I did to change it. 

I've been messing with transient, etc.

I think the renaming works for under all circumstances and the wp-cron stuff seems to work.

Can you please update to latest `wp-cron` branch and test. Thanks a bunch.

afragen:
> I think the latest commit fixes the remaining issues, except I'm now seeing something strange. When doing an AJAX update from the plugins.php page the version number seems be replaced with an empty string. This used to work and I'm not sure what I did to change it.

Odd thing is that it works locally, just not on the server. Testing with 4.3RC (which I've been using all along)

afragen:
> I think the latest commit fixes the remaining issues, except I'm now seeing something strange. When doing an AJAX update from the plugins.php page the version number seems be replaced with an empty string. This used to work and I'm not sure what I did to change it.
> Odd thing is that it works locally, just not on the server. Testing with 4.3RC (which I've been using all along)

Lots of looking at this and I found that when `upgrader_post_install` is able to do `$wp_filesystem->move()` the file gets moved and the version in the plugin row doesn't update. Old version removed and new version empty. Cosmetic glitch but still functional.

This aspect worked much better, though more difficult to code, using `upgrader_source_selection`.

Also, the downside is that it makes it difficult to support remote update services like iThemes Sync and InfiniteWP when the code loads only in specific areas.

afragen:
I've pushed this all over to `develop`. If you were using the `wp-cron` branch just update it and it will pull you over to `develop`.

Let me know how you like it and thanks a bunch.

khromov:
Awesome stuff @afragen 

Been a busy week so far but I'll make sure to extensively test the develop branch some time this week. Thanks for being patient and of course creating GitHub Updater. :)

afragen:
Thanks @khromov. The changes shave off about 0.2 sec of page load. Not tremendous but as a percentage plugin about 2-3x faster.

Believe me I know from busy. :wink: 

Only "bug" I've found is a cosmetic one where during a shiny update of plugin the version number is replaced with an empty value. I've looked into core here and there's no hook available. Odd that it works on localhost but not on server. Update happens correctly.

Not sure it's worth trying to fix or try to ask for hook added to core and fix that way. I'll have to play around.

khromov:
I'm a little qurious about this line:
https://github.com/afragen/github-updater/blob/f06d83e8903cceb20160970bd4f023753786119b/src/GitHub_Updater/Plugin.php#L151

Unless I'm missing something, it means that if the first call to that function looks like this:
```php
Plugin::instance(false);
```

..we're gonna get `(bool)false` back because we won't pass this conditional:
https://github.com/afragen/github-updater/blob/f06d83e8903cceb20160970bd4f023753786119b/src/GitHub_Updater/Plugin.php#L151

Is that intended behaviour?

afragen:
Without that line, if the `Plugin::instance()` hasn't been previously called it will get called with `Plugin::instance(true)`, even if `$force_meta_update = false`.

 That's not what we want as it doesn't short circuit the load. Make sense?

afragen:
I've followed the code flow in xdebug and I believe it results in the behavior that we want.

afragen:
@khromov any other thoughts on this? BTW, it's all now on the `develop` branch.

khromov:
@afragen Great. Just starting my vacation now, and that of course means plenty of coding. :) Will get back to you in the coming days!

afragen:
Have a great vacation.

afragen:
I left a note to move this discussion back to #242. As this PR is closed it should be easier to follow.

PR #275, opened 2015-08-06
Use caret constraint for composer

GaryJones:
See http://blog.madewithlove.be/post/tilde-and-caret-constraints/


afragen:
Thanks Gary. I have know how idea what this means but if you think it needs to be changed -- done.

PR #279, opened 2015-08-15
Refactoring background updates for a cleaner interface

khromov:
The goal of this commit is to refactor how background updates are handled to provide a simpler interface.

* Invoking the constructor of Theme and Plugin classes no longer trigger any kind of updating by itself
* Instead, the functions `remote_plugin_meta_update` and `remote_theme_meta_update` need to be invoked in order to trigger a remote update. This lets us create the classes without having to worry about whether they are updating or not.
* AJAX no longer triggers remote update, ever.
* Broke up forced_meta_update into separate theme/plugin functions

**Known issues**: 
This is a work in progress / suggestion. 

* Plugin class refactoring is done but Theme class still needs some work.
* Branch switcher does not work because it's placed in an odd location, see: https://github.com/afragen/github-updater/pull/279/files#diff-9fc7f6ea61542ac484e16a01eb4d61e1R143

### DIFF
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -169,17 +169,31 @@ class Base {
 	 * Piggyback on built-in update function to get metadata.
 	 */
 	public function background_update() {
-		add_action( 'wp_update_plugins', array( &$this, 'forced_meta_update' ) );
-		add_action( 'wp_update_themes', array( &$this, 'forced_meta_update' ) );
+		add_action( 'wp_update_plugins', array( &$this, 'forced_meta_update_plugins' ) );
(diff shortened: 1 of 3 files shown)


afragen:
@khromov take a look at branch `refactor-constructor`. I did most of the things you we doing here and I don't think anything's broken. Let me know what you think.

https://github.com/afragen/github-updater/commit/6e92e33011decba08162718a5b94b658bd9f5d0d

khromov:
@afragen Code looks neat and clean. I'll give it a spin tonight to test it further, but from a cursory glance it looks very good - great work!

afragen:
Thanks, I think I grok'd what you were going for in this PR. :+1:

PR #290, opened 2015-09-01
csx: fixes error message in Messages.php: 'unknown index page in line…

chrishiggins29:
… 43'. This is b/c $_GET['page'] is not checked if it exists.

### DIFF
--- a/src/GitHub_Updater/Messages.php
+++ b/src/GitHub_Updater/Messages.php
@@ -39,7 +39,7 @@ class Messages extends Base {
 
 		if (
 			! in_array( $pagenow, array_merge( $update_pages, $settings_pages ) ) ||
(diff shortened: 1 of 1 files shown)


afragen:
Can you tell me when this error occurs? I've not seen it.

chrishiggins29:
sure - I always switch on debugging to achieve higher quality, like so:

In wp-config.php

@ini_set( 'log_errors', 'Off' );
@ini_set( 'display_errors', 'On' );
define( 'WP_DEBUG', true );
define( 'WP_DEBUG_LOG', false );
define( 'WP_DEBUG_DISPLAY', true );

Then I see this nasty error since we are not checking if $_GET['page'] even exists (see attached screenie)
githubupdater

afragen:
I see this now. It only appears on the main settings page. Thanks for noticing.

PR #305, opened 2015-09-22
Portuguese (pt_PT) update

pedro-mendonca:


afragen:
There doesn't seem to be much changed.

pedro-mendonca:
That is correct. When you added the context to that string, it got correctly taged as #fuzzy as it might need a check up. I've checked that it's correct and just took out tha #fuzzy tag that doesn't allow it to export to .mo file and to appear translated.

afragen:
Thanks. I added the context after our last commit so others might benefit.

pedro-mendonca:
:+1:

PR #314, opened 2015-10-12
Automatic Updates

nk2580:
Hello All,

I'm looking into the option of including automatic updates and theme hooks to hide the update process from users and keep all updates pushed constantly.

any and all feedback will be greatly appreciated.


afragen:
The goal of this plugin was to approximate the wp dashboard as closely as possible. When WordPress allows for automatic updating of plugins and themes, I'm reasonably certain that it will be supported. 

Currently, @petersuhm's @wp_pusher supports this as does @yafred's wp-alt-repositories .

PR #343, opened 2016-01-05
Add Bulgarian translation

dimitrov-adrian:


afragen:
Thanks @dimitrov-adrian

PR #363, opened 2016-02-20
Do no export bin directory

benoitchantre:
Prevent to install this directory in WordPress


afragen:
Thanks. I must have missed that one.

Issue #354, opened 2016-02-01
RESTful API for remote management


limikael:
I like the remote management feature, but unfortunately I don't use "iThemes Sync" or "InfiniteWP". I'm sure they are great, but I'm kind of stuck in liking my command line. Therefore, it would be great if it would be possible to update a remote site by accessing a RESTful api on the site. That way I can create a script that would update a remote site using curl or something similar. I saw there is some discussion here about supporting wp-cli, but that doesn't really solve my issue, since wp-cli has to be run on the actual server where the site is run. I would like to be able to run an update script to update a remote server.

Please see this as the opening of a dialog. I would be happy to take a stab at creating a PR for this feature. I would like to ask if such a PR would be well received.

And thanks for a great plugin!

afragen:
@limikael I'm all in favor of a PR for REST API endpoints but how would you authenticate to allow for the updates?

Before starting a PR I would really like to know how it would work.

limikael:
Good question!

I envision it like this. On the "Remote Management" tab, there would be a checkbox to enable the API. Next to the checkbox there would be a link that says "show key". The link would only be available to administrator. This link would reveal a randomly generated 24 character key, something like ADRpWzFhMUd5e6NpRemMDEmJ. Maybe there should be a way to regenerate the key and/or set it manually, but not for an MVP. This is a mockup of what the UI could look like:

![github_updater](https://cloud.githubusercontent.com/assets/902911/12742691/39237e12-c999-11e5-8582-e4cc8a7d6e89.png)

Then, the actual api would be available at the following url, and to access it one would have to pass the key to it, like so:

```
<wordpress_base_url>/wp-plugins/github-updater/api.php?key=ADRpWzFhMUd5e6NpRemMDEmJ
```

In the file `api.php` we could not assume that any wordpress cookie was set, of course, and we would not be able to rely on any other "standard" wordpress authentication mechanism. We would however be able to check if the key was passed properly. This check would be the first thing done in that file, and if the correct key was not set we would `die` hard. This was we would ensure that security wasn't compromised by potential sloppy programming further down in the file (or dependencies of the file).

To actually perform an update of a theme or plugins, one would POST to an url like:

```
<wordpress_base_url>/wp-plugins/github-updater/api.php/update?key=ADRpWzFhMUd5e6NpRemMDEmJ&plugin=my/plugin
```

or

```
[...]

PR #366, opened 2016-02-24
Rest api

limikael:
Pull request for the functionality discussed here: https://github.com/afragen/github-updater/issues/354

Short summary:

This PR makes it possible to access a URL in a RESTful way, and as a result update a single theme or plugin. The status of the update will come back as a JSON object. The URL to access is shown on the settings page, under the "Remote Management" tab. The parameters accepted by the URL:

* `key` - The key as displayed on the "Remote Management". Must match the key stored on the
  system.
* `plugin` - Specify this to update a particular plugin. Should be specified as only the name of the
  plugin as it is known to the system, not the remote repo.
* `theme` - Specify this to update a theme.
* `tag` - Specify a particular tag, branch or commit for the update. If nothing is specified, it defaults to
  "master".
* `committish` - This is an alias for the tag parameter. I think that referring to a "committish" is more
  correct in in terms of git terminology, but it sounds a bit funny so this is why I decided to treat the
  committish and tag parameters as aliases for each other.

### DIFF
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -191,54 +191,8 @@ class Base {
 	 * Ajax endpoint for rest updates.
 	 */
 	public function ajax_update() {
-		try {
-			$json_encode_flags=0;
-			if (defined("JSON_PRETTY_PRINT"))
-				$json_encode_flags=JSON_PRETTY_PRINT;
-
-			if (!isset($_REQUEST["key"]) || $_REQUEST["key"]!=get_site_option('github_updater_api_key'))
- 				throw new \Exception("Bad api key.");
-
-			$tag="master";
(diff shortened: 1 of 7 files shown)


afragen:
Thanks @limikael it will take me a while to work through this.

afragen:
I'm going to merge this and you'll have to take me through how to test it later :wink:

I will need to change a number of things to conform to WP Coding Guidelines and this project's "guidelines", but I'll do that at the end. Most of this involves renaming the classes and spacing.

limikael:
Awesome! I see that it is merged into the rest-api branch. What is needed for it to become "official" and released? It would be very nice for us to have it in the mainline so we can deploy it.

afragen:
I've made a few small naming changes. Can you confirm latest `feature/rest-api` works and I'll merge into develop and then `master` at next release.

Issue #383, opened 2016-04-11
Does the Github Updater use WordPress's HTTP API for the auto-updates?


ArchitectforBusiness:
Just checking so I can see if it should be compatible with MainWP's WordPress site management plugin.

afragen:
It is entirely hooked into WordPress core update functions. You might need to check one of the other options under the Remote Management tab. 

If it works with MainWP please let me know and let me know what settings you used so I may officially support it.

ArchitectforBusiness:
Hi Andy,

I reached out to MainWP and got this response:

The issue that we are running into is that the Github Updater is the plugin detecting the updates on your premium. MainWP can not read the update from Github Updater.  

More then likely a specific Extension will need to be made for Github Updater.

Does this sound accurate? Thanks for your help.

Steven Walsh
Architect for Business





architectforbusiness.com
steve@architectforbusiness.com
(860) 510-9328

afragen:
GitHub Updater detects its updates and adds them into the appropriate WordPress option. This is exactly how WordPress itself works. 

InfiniteWP, iThemes Sync, and ManageWP are all able to pick up the updates. The only exception is ManageWP doesn't "see" updates to child themes. 

I haven't looked at MainWP but they must pick up the site_update_transient before GitHub Updater has a chance to run. I don't really know.

afragen:
MainWP support now working but needs next update to MainWP child plugin.

PR #384, opened 2016-04-12
Support MainWP

ruben-:
MainWP uses admin-ajax.php to hit the child sites, this is not yet supported in github-updater. I added this page to the allowed pages for a meta refresh and also added MainWP as a supported remote management system.

This is in reply to ticket #383.


afragen:
Does this make it work?

ruben-:
In combination with our new child plugin that will be released shortly it works. (I'm the head developer of MainWP)

afragen:
Thanks @ruben- I'm so glad to have you're helping with this. I really want it to support everyone.

PR #400, opened 2016-05-25
Documentation for the RESTapi

limikael:
Wrote some documentation for the RESTapi update service. Please check if it is understandable. Also if it is too talkative feel free to shorten it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,17 +280,17 @@ Please go the Remote Management tab of the Settings page and check which remote 
 
 ![Remote Management Tab](./assets/screenshot-3.png)
 
-### RESTapi for Remote Management
+### REST API for Remote Management
 
-GitHub updater also supports other customized continous integration workflows, and it is possible to integrate with other services than those discussed above. For this, there is a RESTapi available for telling GitHub Updater to update themes and plugins to the latest version from their repositories.
+GitHub Updater also supports other customized continuous integration workflows, and it is possible to integrate with other services than those discussed above. For this, there is a REST API available for telling GitHub Updater to update themes and plugins to the latest version from their repositories.
 
-On the Remote Management tab, you will see a url that serves as the enpoint for this api. This url will look something like:
+On the Remote Management tab, you will see a url that serves as the enpoint for this api. This url will look something like this:
 
     http://localhost/wordpress/wp-admin/admin-ajax.php?action=github-updater-update&key=76bb2b7c819c36ee37292b6978a4ad61
 
-The exact url will of course depend on your system. The part after the `key` attribute is automatically generated on the first activation of the GitHub updater plugin and is used for authentication. Any person or entity knowing this key will be able to change the versions of your installed plugins, but nothing else.
+The exact url will of course depend on your system. The value for the `key` attribute is automatically generated on the first activation of the GitHub Updater plugin and is used for authentication. Any person or entity knowing this key will be able to change the versions of your installed plugins, but nothing else.
 
-Now, if we would use `curl` to access the url exactly like it is written on the Remote Management tab, we would see something like this:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -280,6 +280,49 @@ Please go the Remote Management tab of the Settings page and check which remote
...
+
+    $ curl "http://localhost/wordpress/wp-admin/admin-ajax.php?action=github-updater-update&key=76bb2b7c819c36ee37292b6978a4ad61"
+    {
+        "message": "No plugin or theme specified for update.",
+        "error": true
+    }
+
+This error message is given because GitHub updater requires us to specify either a theme or a plugin that we wish to update. This is specified using the `theme` or `plugin` attributes, and the theme or plugin is identified by its slug. Let's try to update a plugin:
+
+    $ curl "http://localhost/wordpress/wp-admin/admin-ajax.php?action=github-updater-update&key=76bb2b7c819c36ee37292b6978a4ad61&plugin=myplugin"
```
afragen: FYI, people will use this to test. Something to consider as you must now keep this repo and not delete it. If you want we can switch it to one of my test plugins.

README.md
```diff
@@ -280,6 +280,49 @@ Please go the Remote Management tab of the Settings page and check which remote
...
+            "Downloading update from <span class=\"code\">https:\/\/api.github.com\/repos\/limikael\/myplugin\/zipball\/master<\/span>&#8230;",
+            "Unpacking the update&#8230;",
+            "Installing the latest version&#8230;",
+            "Removing the old version of the plugin&#8230;",
+            "Plugin updated successfully."
+        ],
+        "success": true
+    }
+
+And our plugin is updated! The messages that are shown are those that otherwise would be shown in the Wordpress admin interface. The full list of attributes accepted by the service is shown below:
```
afragen: These messages will no longer be shown in Shiny Updates.

README.md
```diff
@@ -280,6 +280,49 @@ Please go the Remote Management tab of the Settings page and check which remote
 
 ![Remote Management Tab](./assets/screenshot-3.png)
 
+### RESTapi for Remote Management
```
afragen: Is the standard `RESTap` or is it `REST API`? We should try to be consistent with how WP uses the expression.

README.md
```diff
@@ -280,6 +280,49 @@ Please go the Remote Management tab of the Settings page and check which remote
 
 ![Remote Management Tab](./assets/screenshot-3.png)
```
afragen: I will create a new screenshot for this. As it will be for my localhost it really shouldn't matter that it's published.

afragen:
Overall, it looks great. A couple of inline comments for editing.

README.md
limikael: Good point. Actually, the url was pointing to a repo that didn't even exist, I just thought of it as an example. I created a separate repo and pointed it to that one, so now it will work if people copy-paste from the docs. If you don't mind, I would like to keep it pointing to one of my repos. This is partially because I like to see my name there, and partially because of the street cred it would give me from all the hackers (or recruiters) who would see my name. However, you are the main author of the plugin, so if you would like to see your name there for the same reason, I have total respect for that.

README.md
limikael: What is shiny updates? Is it this https://wordpress.org/plugins/shiny-updates/? And how does it relate to github-updater?

README.md
limikael: Seems that here: https://developer.wordpress.com/docs/api/ and in other places they use REST API, so I changed it into that.

README.md
limikael: Ok

limikael:
Pushed a new commit with fixes based on your comments, and also some spelling mistake fixes.

PR #424, opened 2016-07-25
Translation update (pt_PT)

pedro-mendonca:


afragen:
As always, thanks Pedro.

PR #426, opened 2016-08-02
Use theme and plugin dirname, if exists, instead of repo name

joelworsham:
If you manually install a plugin/theme, and give it a dirname, which is common, then the updater should use that dirname, rather than searching for the repository slug. In many cases, the repository slug is not what I want the dirname in my site to be.

Example. Say I have an organization named "Cool Stuff" and I have a plugin in that organization named "plugin-a". I may want the site this applies to have the plugin installed in "cool-stuff-plugin-a", rather than just "plugin-a".

Anyways, I don't know that I covered all grounds here with the refactor, but I think I'm close. And that's all this PR is, a refactor of the usages of the `$theme->repo` and replacing with `$theme->slug`, basically. Same for plugins.

### DIFF
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -537,7 +537,7 @@ class Base {
 		 */
 		if ( $this instanceof Plugin &&
 		     ( defined( 'GITHUB_UPDATER_EXTENDED_NAMING' ) && GITHUB_UPDATER_EXTENDED_NAMING ) &&
-		     ( ! $this->config[ $repo['repo'] ]->dot_org ||
+		     ( ! $this->config[ $slug ]->dot_org ||
 		       ( $this->tag && 'master' !== $this->tag ) )
 		) {
 			$new_source = trailingslashit( $remote_source ) . $repo['extended_repo'];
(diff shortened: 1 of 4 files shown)


afragen:
Fundamentally this plugin is designed to mimic the WordPress core updating experience when your plugin is hosted outside the w.org repo.

The plugin repo has specific naming guidelines to avoid conflicts. GitHub already gives the dev the ability to name their repo. GitHub Updater already has a mechanism to cope with the small possibility of duplicate repo names from different developers.

Trying to troubleshoot why this plugin may not be working for someone who has an arbitrary plugin dirname is a support nightmare.

This is an interesting request and I do have to say it's the first time I've had such a request, whether via an issue or PR. Your PR also doesn't address the renaming that must occur for a downloaded update. The arbitrary dirname isn't available to the renaming method.

I'm still not clear on why the arbitrary directory name? The only person looking at it is whomever looks on FTP. There's no functional benefit, is there?

joelworsham:
I see your points. For me, it's simply a semantic naming thing. I always prefix my plugin dirnames with the company/project ID, in order to avoid conflicts. Whereas, on the hosted repo on GitHub, I don't prefix, because it would be redundant given that it exists in the company/project organization.

So, perhaps it's an edge case. In my mind it seemed unobtrusive, given that the default when downloading a plugin/theme through GitHub Updater, it would inherit the repo slug, whereas if the plugin was installed manually, it would use the dirname.

Anyways, just my two cents. This PR didn't take long and it's been handy for me, so I simply figured I'd push it up. I won't be disappointed if you don't merge it.

afragen:
Have you looked at the extended naming option? It's only for plugins.

joelworsham:
I did, and it seems to be a valid solution. My nitpicky self would still prefer a dirname different than that, but I know I'm just being overly picky ;)

PR #430, opened 2016-08-19
If the REST endpoint is accessed from a GitHub Webhook, pick up the tag from the payload

limikael:
I noticed that when the REST endpoint is used in a GitHub webhook, there might be a race conditions some times when the zip file is downloaded as e.g. `.../zipball/master`. This PR checks if the pushed commit is the same as the specified tag, and if so it changes the url to be explicit to the commit, like `.../zipball/726b61d522e90aa0b89764a0d53ed0763a42b87c`. This seems to take care of the race condition.

Btw, I have only tested with GitHub, I don't know if there are similar problems using e.g. BitBucket.


afragen:
I don't think I've ever encountered this, but not sure I would know it if I did encounter it.

I thought GitHub zip files were created on the fly.

I'm. It sure why you would need to check if the incoming event is from a webhook as that's the only way this code is reached.

Have you encountered a race condition? What happened?

limikael:
Hi @afragen!

Thanks for the quick reply, as always! :)

Yes I also think they are created on the fly, more or less. However, there is probably some caching involved and maybe clustered servers and such. I don't know this for sure, but it would make sense to me that things at GitHub's end would be built that way, in order to not have to generate the zip file each time, since this would be quite computationally expensive.

I have encountered this condition, so the PR came from solving a real problem I had, it was not just speculation. What happened was that at maybe 20% of the times or so, the update did not pick up the latest version but the previous one. I.e. it "updated" to the same version that was already deployed, not the new version that was just pushed. If I afterwards would wait for 30 secs and then hit the REST update url manually using curl, then the new version would be picked up. From this I drew the conclusion that there was some cache somewhere that wasn't invalidated or something. When using a version of `github-updater` with the code in this PR, the problem went away.

 Well, maybe "checking if the event is from a Webhook" is the wrong formulation. The code I was referring to was this:

```
$hook_data = $this->get_github_webhook_data();
if ($hook_data && $tag == $hook_data["branch"]) {
    $tag = $hook_data["hash"];
}
```

I.e., we check if the data from the Webhook is there, if it is we use it. We can't rely on it just being there always, as the REST update url could be hit by manually curl or by some other system, and we don't want it to mysteriously crash if the expected Webhook parameters aren't there.

Does this make sense?

Btw, another thing... I noticed that when using the 5.5 release version and the code currently in `master`, then the reactivation of the plugin after the update didn't work. When using the code from the `develop` branch the reactivation did work. I assume that this is a known and fixed problem and all I need to do is to wait for the code from the `develop` branch to be merged into `master` and eventually released, but just checking?

afragen:
> Btw, another thing... I noticed that when using the 5.5 release version and the code currently in master, then the reactivation of the plugin after the update didn't work. When using the code from the develop branch the reactivation did work. I assume that this is a known and fixed problem and all I need to do is to wait for the code from the develop branch to be merged into master and eventually released, but just checking?

I think I encountered this issue and fixed it in the `develop` branch.

As the issue seems to be related to a timing issue, what about introducing a slight time delay before the doing the update? This would likely be much simpler to add and would be more universal should Bitbucket or GitLab have similar issues. What do you think?

limikael:
Hm, I don't think that introducing a time delay would be the right way to go. It would degrade performance (I.e. it would take longer to do the update) and it wouldn't really be possible to get it to work reliably. Regardless of how long the time delay would be, there is always some kind of chance that the cache still wouldn't have been invalidated in a correct way. Unless the time delay is an hour or so, but that would degrade performance to the point where it would be almost useless... :)

Introducing time delays and such would to me seem like a hack that could be used as a last resort, when no other way is available. Doing it like it is currently done in the PR uses the data from the payload from BitBucket according to the way it is documented and intended to be used. To me this is more reliable. 

It seems that you are not quite happy with the way the PR works. For the sake of the discussion, could you maybe elaborate a bit on the reasons for this? Is it because it is GitHub specific? If that's the case I think we could implement a similar thing for other supported systems.

afragen:
@limikael the only thing I wasn't fond of was a GitHub only process. In all fairness, this is probably wrong on my part.

If I understand the PR correctly it essentially takes `&tag={branch}` and converts it to `&tag={commitish}`. As all of our RESTful endpoints are the same, we should ensure that this change is across all platforms. And we should test. :wink: I'll create a branch for the PR to go against instead of develop.

I assume this change isn't needed for `&tag={tag_version}` or should we try to convert every non-commitish endpoint ( `tag` ) to a commitish endpoint ( `commitish` )?

I'd like to test this more and firm up all instances and locations for implementation. The other thing is, eventually I'd like to create an actual REST API endpoint for this and want to ensure it will work in that instance too.

limikael:
Yes that's pretty much what it does, changing `&tag={branch}` to `&tag={commitish}`. It does it in a conditional way though, and it only makes this change if the incoming commit as specified in the payload from GitHub matches the branch specified in the URL.

Hm... I don't think it would make sense for `&tag={tag_version}`, it would only make sense for branches. Because, I think that a `tag_version` is not something that you commit you code to, right? I mean you commit the code _to_ a branch. A version is a snapshot of the code at a particular time, but you can't commit code _to_ the version? Or can you?

Yes some refactoring and changing how things work and making it more like an actual REST API would be nice! However, as you say, before this point it would be nice to have things really tested. This also means "testing" the documentation, i.e. do people understand how to use this? I'm planning to write an article about this. We can see people's reactions to the article as a kind of test and validation on if the idea is right, and then we can refactor things and make it more "proper" later.

limikael:
And I assume that since it is a new branch it needs a new PR? I created one here:

https://github.com/afragen/github-updater/pull/433

afragen:
> Hm... I don't think it would make sense for &tag={tag_version}, it would only make sense for branches. Because, I think that a tag_version is not something that you commit you code to, right? I mean you commit the code to a branch. A version is a snapshot of the code at a particular time, but you can't commit code to the version? Or can you?

You are correct and yes a tag/release is a point in time and shouldn't need this mod.

I think GitHub now allows you to change the branch on a PR here.

afragen:
#433 merged. I need to fix something with private Bitbucket updates and issues with shiny updates. I have a solution but not ideal. Once I get that out I'll test this code and merge into develop.

Thanks @limikael !

PR #433, opened 2016-08-24
Restful commitish endpoints

limikael:
Creating a new PR after the discussion here:

https://github.com/afragen/github-updater/pull/430#issuecomment-241283362


limikael:
I'm testing the bitbucket API as well to see if the same race condition / timing issues exists there... Thought I would be fast enough to maybe put that in the same PR, but I'll create a new one... Keep the branch alive for just a little... :)

afragen:
Will do.

limikael:
I have verified that similar timing issues exist when using BitBucket. The webhook sends the data needed to fix it though, so it should be ok! The data is on another format than from GitHub, so there needs to be a similar but slightly different fix to handle it. Don't have time exactly now to fix it, but it will come within the next couple of days!

PR #435, opened 2016-08-26
Restful commitish endpoints

limikael:
Fixed it for Bitbucket also...

Parsing the webhook data wasn't 100% as straightforward as it was to parse the data from github. Specifically, I asked a question here:

http://stackoverflow.com/questions/39165255/how-do-i-get-the-latest-hash-from-a-bitbucket-push-payload-why-is-changes-an

But it seems to work!

Ops... Btw, I see that the commit message says "parse github webhook payload". That was a typo, it should say "parse bitbucket webhook payload" of course.

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,7 +323,8 @@ And our plugin is updated! The messages displayed are those that otherwise would
 
 When using the RESTful endpoints for updating themes or plugins, you need to specify at least the `key` attribute, as well as one of the attributes `plugin`, `theme`, or `updates`. All other attributes are optional.
 
-If the incoming request looks like it is push event from a [GitHub Webhook](https://developer.github.com/v3/activity/events/types/#pushevent), and if the pushed branch matches what is specified in the `tag` attribute, then the update will be made according to the latest push as specified in the event. This is done in order to avoid race conditions where the zip file for the branch might not have been created yet on GitHub.
+The RESTful api is useful, among other things, for automatically updating themes and plugins on events sent as Webhooks from GitHub and the other services supported by this plugin. Some special functionality has been implemented to support this in order avoid race conditions, i.e. to make sure that the updated version is really the version that was just pushed to the repository. Specifically, what this functionality does is to see if the incoming request looks like it comes from a 
+[GitHub Webhook](https://developer.github.com/v3/activity/events/types/#pushevent) or a [Bitbucket Webhook](https://confluence.atlassian.com/bitbucket/event-payloads-740262817.html#EventPayloads-Push). If this is the case, and if the branch that was pushed to matches the branch specified in the `tag` attribute, then the update will be made according to the latest push as specified in the event. Note that this is currently implemented for GitHub and Bitbucket only.
 
 ## Extended Naming
 
--- a/src/GitHub_Updater/Rest_Update.php
+++ b/src/GitHub_Updater/Rest_Update.php
@@ -228,6 +228,52 @@ class Rest_Update extends Base {
 	}
 
 	/**
+	 * See if a tag came in through a bitbucket webhook. It returns data
+	 * on the same format as get_github_webhook_data.
+	 *
+	 * We assume here that changes contains one single entry, not sure if
+	 * this is a safe assumption:
+	 *
+	 * http://stackoverflow.com/questions/39165255/how-do-i-get-the-latest-hash-from-a-bitbucket-push-payload-why-is-changes-an
+	 */
+	private function get_bitbucket_webhook_data() {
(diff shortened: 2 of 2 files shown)


afragen:
So @limikael should I ask if GitLab webhooks are any different?

afragen:
Given the similarity of the code, is there a way to pass whether the source is from GitHub, Bitbucket, or GitLab and create a single function?

I'm not sure how to test this code, can you share how you test this?

limikael:
These are the steps I used to test it. Might be more details than you need, but better with too much detail rather than too little.

1. Create a Wrodpress site that's easy to deploy to. I used http://www.cloudaccess.net/ since it is free.
2. Zip the working copy of github-updater and install it.
3. Use github-updater to install a plugin from GitHub. I used my plugin here: https://github.com/limikael/mickesplugin
4. In GitHub settings for the test plugin, under "Webhooks & services", you will find the place to create a webhook. Click "Add webhook".
5. In the "Payload URL" for the webhook is where you enter the URL from the "Remote Management" tab. You need to add the plugin name to the url, so my URL was http://cdtest.cloudaccess.host/wp-admin/admin-ajax.php?action=github-updater-update&key=a1dd569f2b807a4a9241c1f3dccbcc06&tag=master&plugin=mickesplugin
6. Now, push to the repository for the test plugin. You should then see the plugin being updated on the site automatically. This test worked before 80% of the times, but failed about 20% of the times before the PR. With the PR it should work more reliably.
7. If you go in and click "Edit" next to the Webhook, you will see a section at the bottom "Recent Deliveries". You should see an entry here corresponding to each push.
8. If you look at the details for the Webhook delivery, you can see the response that the plugin sent back to GitHub.
9. In my case, the response looks something like this: 

    ```
    {
        "messages": [
            "Downloading update from <span class=\"code\">https:\/\/api.github.com\/repos\/limikael\/mickesplugin\/zipball\/15c58d47243ec8849fe81c7553abc19dfff82662<\/span>&#8230;",
            "Unpacking the update&#8230;",
            "Installing the latest version&#8230;",
            "Removing the old version of the plugin&#8230;",
            "Plugin updated successfully.",
            "Plugin reactivated successfully."
        ],
        "success": true
    }
    ```

10. The interesting thing about the response is the `...mickesplugin/zipball/15c58d47243ec8849fe81c7553abc19dfff82662...` part. It was not like this before the PR, then it said `...mickesplugin/zipball/master...` instead. So here we can verify that the plugin actually changed the tag to the hash.

The steps above are for GitHub, but it works very similar for BitBucket. BitBucket also has the possibility to view the log of delivered Webhooks and the responses can be inspected for debugging.

Makes sense?

limikael:
Regarding GitLab, yes I assume that it would be possible to create a similar fix.

You are also right that this could be moved to a single function. I would actually suggest even moving it to its own class. We could create a class called WebhookParser that parses these responses for the different services, and then just have a single call from within the Rest_Update class. This class could then also hopefully be reused if we refactor things later.

Will start working a bit on this WebhookParser class and also maybe see if I can get the class to parse GitLab Webhook calls.

If you think this seems like a good architecture?

limikael:
Ah another thing regarding testing... 

What I described above was how I tested it once I had a developed candidate to test. Before this I copy pasted a request payload from GitHub/Bitbucket and saved in a local file. I used the file and sent it to localhost using curl with a command similar to `curl -X POST -d @attic/bitbucketpayload.json "http://localhost/wordpress/wp-admin/admin-ajax.php?action=github-updater-update&key=ffa9bd3f2b244496fc1bc20fd2ae1f1c&plugin=mickesplugin&tag=master"`. This is useful during development, and also gives a chance to use the `error_log` function for tracing out info.

And maybe the idea with a separate class was overkill. But this code is duplicated in both functions:

```
		$request_body = file_get_contents('php://input');
		$request_data = json_decode($request_body, TRUE);
```

I will move it to a single function.

afragen:
If you think a separate class is better that's fine. It just seems like the separate class would only have a single function. 😉 

Is there something in the payload that says it's from GitHub v. Bitbucket? I ask as this would the simplest way to parse the different parsing schemes.

Regarding testing. I was looking for how to test locally, but I guess that's not really feasible.

afragen:
Is it possible to get the webhook headers? This might make it simpler to determine the source of the webhook.

limikael:
Well it is sort of feasible to test locally, using curl as I wrote about in a previous comment.

Yes, the Webhooks do send headers to identify them. This is probably a cleaner approach, I will refactor the code to use the headers instead.

PR #436, opened 2016-08-27
Check headers from request to determine if it came from a webhook

limikael:
As discussed here: https://github.com/afragen/github-updater/pull/435

I changed the webhook code to check the http request headers to see if the request came from a webhook on GitHub or Bitbucket. Previously various parts of the request body were checked instead and I attempted to "guess". Relying on the headers is a cleaner solution.

Also this resulted in a cleanup of the code. Since before we needed to make a "guess" we needed to be very pragmatic if the request came on the right format. Not we can just assume that if the headers say that it is a GitHub Webhook, then we can parse it as a GitHub Webhook, and we don't need so much error checking.

Will take a look at GitLab Webhooks also next.


limikael:
Made it parse GitLab Webhooks also.

It seems however, that GitLab doesn't suffer from the same race conditions as GitHub and Bitbucket, so this might not be strictly necessary. But I think it's good to parse the request and act on it in the same way for the sake of consistency.

src/GitHub_Updater/Rest_Update.php
```diff
@@ -240,42 +253,70 @@ private function get_bitbucket_webhook_data() {
...
+	 * We need to rely on the latest commited hash from the remote repository
+	 * and be explicit when specifying the tag we want to update to. If we
+	 * don't do this there is a chance for a race condition, since the default
+	 * zip file on the repository service might not have been created yet.
+	 */
+	private function get_webhook_data() {
+
+		// GitHub
+		if ( $_SERVER["HTTP_X_GITHUB_EVENT"] == "push" ) {
+			return $this->get_github_webhook_data();
```
afragen: We should rename this function to `id_webhook()`

Then we could have a single method for parsing and pass `$results_data` to smaller functions for each service.

The flow would look something like.

```php
function process_request()
    $this->get_webhook_data();

        $webhook_source = $this->id_webhook;
        $this->parse_webhook_data( $webhook_id, $request_data );

private function parse_webhook_data( $source, $data ) {
    if ( 'gihub' === $source ) {
        $this->parse_github_webhook( $data );
    }
}

or something like that.

In this way we can further simplify each function and keep them with a single responsibility.

Thoughts?

afragen:
We can clean it up later.

afragen:
It's later 😉 

https://github.com/afragen/github-updater/commit/af9443f3ecadaa0154610a5e529cda2fb02db923
https://github.com/afragen/github-updater/commit/abf3ec2a3801069f04ce06496b17a4468533846a
https://github.com/afragen/github-updater/commit/b948ff12b70e5d8d3f450ed673bedd62671015c7

You will need to rebase against the upstream branch to test.

limikael:
Awesome @afragen! I gave it a smoketest (deployed your new version to my site and did a test update of my testplugin). Seems to work!

afragen:
Merged https://github.com/afragen/github-updater/commit/6481b4fe5f4d523cc7d80cc65eee4ef23c42c5cf

PR #455, opened 2016-09-26
Needs PHP Zip extension.

nextgenthemes:
WordPress Version: 4.6.1
PHP Version:       7.0.11-1+deb.sury.org~xenial+1

on Ubuntu 16.04 and I had the plugin failing with 

> The package could not be installed. PCLZIP_ERR_BAD_FORMAT (-10) : Invalid archive structure

until I installed `php-zip`. Maybe there could be a better error message because is was just a guess and the message itself does not really point to the solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 * Contributors: [Andy Fragen](https://github.com/afragen), [Gary Jones](https://github.com/GaryJones), [Seth Carstens](https://github.com/scarstens), [Mikael Lindqvist](https://github.com/limikael), [contributors](https://github.com/afragen/github-updater/graphs/contributors)
 * Tags: plugin, theme, update, updater, github, bitbucket, gitlab, remote install
 * Requires at least: 4.0
-* Requires PHP: 5.3
+* Requires PHP: 5.3 (and `php-zip`)
 * Tested up to: 4.6
 * Stable tag: master
 * Donate link: http://thefragens.com/github-updater-donate


afragen:
What are you trying to do that's causing this issue? I've seen similar errors when a private repo is trying to be installed/updated and the credentials are incorrect.

There is no requirement for PHP Zip.

afragen:
Processing a zip file is how WordPress core functions. Look at uploading a plugin for installation.

nextgenthemes:
Well thats funny because installing php-zip solved it for me. I was trying to update plugins with the github updater. WP updates went well but everything from github failed with this message, last was the updater itself but i had the same error with the title capitalization plugin.

Strangely enough i thought I have already tried this and was about to file a issue, just want to make sure i tried with and without php-zip and it was not installed so I did install it and then the updated went through.

afragen:
Does your phpinfo() show no installed zip?

afragen:
That type of error has almost always been an authentication issue.

nextgenthemes:
Not sure why you ask me about if phpinfo() shows installed zip now but I am pretty sure it will show it as installed now since I just installed it. Maybe its all in my head and just random but I thought the php-zip did the trick.

I have no clue about how core handles things but my guess was something like:

Core uses some class for .zip but when it detects php-zip installed it uses that instead. maybe the github zips are a format only the real thing handles well but the class fails with.

Trying something ...

afragen:
> Not sure why you ask me about if phpinfo() shows installed zip now but I am pretty sure it will show it as installed now since I just installed it. Maybe its all in my head and just random but I thought the php-zip did the trick.

I ask because wondering if there are now 2 types of zip installed. It would be difficult to imagine the server didn't have a php zip installed.

> I have no clue about how core handles things but my guess was something like:
> 
> Core uses some class for .zip but when it detects php-zip installed it uses that instead. maybe the github zips are a format only the real thing handles well but the class fails with.

I haven't found github's zips to be different.

If you can give details and I'm able to replicate the issue it would help.

nextgenthemes:
OK uninstalled php-zip changed version number of the github-updater to force update. Same error, tried again update went though without zip. Looks like I have to do update 2 times now for whatever reason.

So you're right, had nothing to do with php-zip being installed.

nextgenthemes:
> 2 types of zip installed

?? Not sure what you mean but my "server" is my home PC with ubuntu 16.04 and the package is called php-zip and WP does not need it as you pointed out. So no clue what you mean with 2 types of zip installed now.

afragen:
What I meant was, possibly there was already a zip installed, though if it was the same module it would have just overwritten.

Here's what I have in my local Docker setup, in Pressmatic.

screenshot_01

afragen:
Was updating previously working on this PC?

afragen:
> OK uninstalled php-zip changed version number of the github-updater to force update. Same error, tried again update went though without zip. Looks like I have to do update 2 times now for whatever reason.

You could have used the branch switch feature to reinstall any branch of any plugin/theme. It will perform a complete re-installation.

> So you're right, had nothing to do with php-zip being installed.

Glad it seems to be working now.

nextgenthemes:
lol I should actually check for `php7.0-zip` and that was installed the entire time. Just uninstalled `php7.0-zip` and forece update and it went though on the first try. Seems some random thing.

PR #501, opened 2016-11-21
Filtered options

teolaz1990:
filtered global options to hook and filter them using plugins

### DIFF
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -131,13 +131,37 @@ class Base {
 	 * @var bool
 	 */
 	protected static $load_repo_meta;
+	
+	/**
+	 * @var string[] Option keys refer to option names, option names are used to save value on database
+	 */
+	//TODO: these should be used on any other part where setting options are used (i.e. the Settings class)
+	protected static $options_names = array(
+		'github_access_token' => 'github_access_token',
+		'bitbucket_username' => 'bitbucket_username',
+		'bitbucket_password' => 'bitbucket_password',
+		'gitlab_access_token' => 'gitlab_access_token',
+		'gitlab_enterprise_token' => 'gitlab_enterprise_token',
+		'branch_switch' => 'branch_switch',
+	);
+	
+	/**
+	 * @var string[] Option keys refer to option names, option names are used to save value on database
+	 */
+	//TODO: these should be used on any other part where setting options are used (i.e. the Settings class)
+	protected static $options_remote_names = array(
+		'ithemes_sync' => 'ithemes_sync',
+		'infinitewp'   => 'infinitewp',
+		'managewp'     => 'managewp',
+		'mainwp'       => 'mainwp',
+	);
 
 	/**
 	 * Constructor.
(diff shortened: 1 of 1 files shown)


afragen:
Other than being able to set the remote installation service programmatically, what's the point? The current hook provides a method to set this data. Have you tried it?

If you're looking for another way for remote management take a look at https://github.com/afragen/github-updater/wiki/Remote-Management

teolaz1990:
These hooks are more ordered than before IMHO and allows you to filter every global option singularly, without messing up with the options array. And, they ensure you and us that if in 2 years you'll have to change $options key array, you won't have any problem as you decided options name directly on $options_names and $options_remote_names

afragen:
Can you provide code on how this would be used?

teolaz1990:
Of course:

on functions.php
`<?php 
   add_filter('github_updater_filter_option_bitbucket_username',function($value){
       return 'myprivateusername';
   });
?>
`

afragen:
And this is different than

```php
add_filter( 'github_updater_token_distribution', function() {
    return array( 
                'bitbucket_username' => 'myprivateusername',
                'bitbucket_password' => 'mypassword',
                'private-theme-slug' => '1',
                 );
});
```

teolaz1990:
Maybe you haven't read what i wrote before... please read better.

afragen:
So a bit of historical perspective. This function has been asked for/about less than a dozen times. What you haven't told me is why the current hook is inadequate. What I see is that with this PR multiple  `add_filter` calls will be needed to set changes.

If your issue is with the naming of the filter we can discuss that using a deprecated notice.

If the wiki is unclear about what is possible, let's fix it.

afragen:
It would be simple enough to rename the hook, see branch `deprecate-token-hook`. https://github.com/afragen/github-updater/commit/376094dc939fbd73955be09bfd4a3dab3ff6b96f

afragen:
FYI, hook has been renamed `github_updater_set_options` https://github.com/afragen/github-updater/commit/f64cef6051a7f9f2b248035fd3bda29028e7c3fc

teolaz1990:
Good, you've done a more descriptive hook now.
What i meant before is another thing... And just to be sure you won't do what i have in mind, i would ask you to put all the available filterable options on documentation:

GLOBAL FILTERABLE OPTIONS:
github_access_token (string)
bitbucket_username (string)
bitbucket_password (string)
gitlab_access_token (string)
gitlab_enterprise_token (string)
branch_switch (bool)

PER REPO FILTERABLE OPTIONS (could appear multiple times)
mytheme (bool|string) (where mytheme is the slug of the theme hosted remotely, bool in case of a private bitbucket repo, string in case of a github/gitlab hash)
myplugin (bool|string) (where myplugin is the slug of the plugin hosted remotely, bool in case of a private bitbucket repo, string in case of a github/gitlab hash)

This thing allow us to BE PROTECTED in case you in a remote time forget to add/remove a filter from documentation and our system starts to work badly.

teolaz1990:
If you want, close this PR.

PR #563, opened 2017-06-24
Treat gitlab group as the repo owner

rolandsaven:
Hi,

I've made some changes to my local copy of the plugin to support GitLab's grouping feature not knowing that there is already a branch for this. 

I'm still submitting my pull request to see if we can learn from it. To my understanding in GitLab, a group is the actual owner of the repository. We should then treat it that way and avoid complication and new array elements/variables.

Let me know what do you think. I'm happy how this branch works, I'm just throwing this out there as an idea to consider.

### DIFF
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -840,37 +840,16 @@ class Base {
 	 *
 	 * @param string $repo_header
 	 *
-	 * @return array $header
+	 * @return array
 	 */
 	protected function parse_header_uri( $repo_header ) {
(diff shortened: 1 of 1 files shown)


afragen:
@rolandsaven this is a great concept, thinking of the group as an extended owner. I've borrowed the concept and like in this PR have greatly simplified `parse_header_uri()`, https://github.com/afragen/github-updater/commit/22e4df3c87eb33d7b387573832e2f1f4112946d3

I'm going to close this PR as I think I've accomplished the goal here.

PR #587, opened 2017-08-17
delete active theme cache during load_hooks()

### LINKED ISSUE #586: Bitbucket Private Repo Themes + Active Theme Issue
I apologize for not providing more details, but have just lost multiple hours to this issue and want to just document it and get back to work.

If you delete the folder via FTP for the active WP theme, then attempt to install that theme using the GHU plugin, specifically I was trying to install a private Bitbucket Repo, it will never appear in the list of installed bitbucket themes on the GHU bitbucket tab (very likely the bitbucket tab itself will never appear).

Once I switched to a different theme, and installed the private repo theme worked fine.

bradmkjr:
closes #586 with minimal performance impact for all users.

Any cache made prior to this point would be lacking extra headers required by GHU, so best to clear cache and ensure fresh headers are generated after plugin is initialized.

### DIFF
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -228,6 +228,24 @@ class Base {
 		add_filter( 'upgrader_source_selection', array( &$this, 'upgrader_source_selection' ), 10, 4 );
 
 		/*
+		 * The following clears any active theme cache, which may have be set by a mu-plugin
+		 * calling wp_get_theme()
+		 */
+		if ( empty( $stylesheet ) )
+			$stylesheet = get_stylesheet();
+		
+		if ( empty( $theme_root ) ) {
+			$theme_root = get_raw_theme_root( $stylesheet );
+			if ( false === $theme_root )
+				$theme_root = WP_CONTENT_DIR . '/themes';
+			elseif ( ! in_array( $theme_root, (array) $wp_theme_directories ) )
+				$theme_root = WP_CONTENT_DIR . $theme_root;
+		}
+		
+		$cache_hash = md5( $theme_root . '/' . $stylesheet );
+		wp_cache_delete( 'theme-' . $cache_hash, 'themes' );
+
+		/*
 		 * The following hook needed to ensure transient is reset correctly after
 		 * shiny updates.
 		 */


afragen:
I’ll have to look closely at this. I do recall that deleting or flushing the cache can be an expensive process for those with Redis caches.

bradmkjr:
An alternative would be to set the cache value to a non array, or false value, that maybe easier for redis, I don't know. 

class-wp-theme.php

```
		$cache = $this->cache_get( 'theme' );
		
		if ( is_array( $cache ) ) {
```

afragen:
I wonder if `wp_clean_themes_cache()` work?

bradmkjr:
It's a judgement call on your part, the advantage to using wp_clean_theme_cache is that it's  WordPress best practice and likely won't break without warning in the future, the major disadvantage is the performance, looking at the code it clears the cache for every theme in the theme directory, that could be 20 or 30 themes (we all have dealt with those sites), and that would run on every instance of WordPress initializing. 

https://core.trac.wordpress.org/browser/tags/4.8/src/wp-includes/theme.php#L114

It also because it scans the theme directory requires server hard drive access. Calls the following function, which does cache the themes for 30 minutes but still I think should be avoided.

https://core.trac.wordpress.org/browser/tags/4.8/src/wp-includes/theme.php#L401

The bug is only the result of the active theme being cached, so I would side with either deleting or overwriting with a non array value to the active theme cache, instead of purging the cache for the entire directory.

afragen:
If we just want to delete the cache of the current theme, what about

```php
$theme = wp_get_theme();
$theme->cache_delete();
```

I haven’t tried it yet.

bradmkjr:
It would functionally work, but it generates the cache when you call
wp_get_theme, and then it also generates a new cache at the end of the
cache_delete because it calls it’s own construct again, so those 2 lines
write 8 cache entries and delete 4.

There may not be a perfect solution for this, just trying to keep it
efficient as this code will run every time WP initializes, and really only
solves a very small problem for a few people. It maybe better to just write
a few line plugin to patch this issue for people on GoDaddy hosting using
GHU.

Thanks
Brad

afragen:
I like that idea.

afragen:
Closed in favor of https://github.com/afragen/github-updater/commit/31ed7061d8667263f010877cb5554b7ea752f3af

Issue #607, opened 2017-10-02
Unable to update private Bitbucket repo with WP-CLI


afragen:
I think it's an issue passing the Basic Auth header. I have to dig deeper.

maremmano:
Hello,

I get the following error when trying to update a theme from private bitbucket repo using wp-cli:
PCLZIP_ERR_BAD_FORMAT (-10) : Unable to find End of Central Dir Record signature

No problem updating from wp admin.

Thank you very much for the work done.

afragen:
That is the error I’m describing.

PR #616, opened 2017-10-24
Update Basic_Auth_Loader.php

v8-ict:
Bugfix for https://github.com/afragen/github-updater/issues/607

If the BasicAuthLoader will be called from wp cli "wp plugin update <pluginname>"
$slug remains empty, and it returns an object of configs, created by the previous Singleton::get_instance( 'Plugin' )->get_plugin_configs(),

This fix makes sure it looks for the matching slug, and returns the type property as mean't in the statement below,
so that it goes into the switch statement correctly, and adds authentication headers.

### DIFF
--- a/src/GitHub_Updater/Basic_Auth_Loader.php
+++ b/src/GitHub_Updater/Basic_Auth_Loader.php
@@ -161,6 +161,17 @@ class Basic_Auth_Loader {
 				Singleton::get_instance( 'Theme' )->get_theme_configs()
 			)
 			: false;
+		
+		if ( ! $slug ) {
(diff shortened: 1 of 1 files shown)


afragen:
@v8-ict thanks so much for finding the issue. I think I found a slightly different, though not necessarily better way. I chose this way as what we really want to do is set `$type`. Otherwise a couple of things changed slightly and rearranged. Can you adjust your PR to the changes in branch `wp-cli` and I will merge so you can have credit?

Seems to work for me. Can you confirm?

v8-ict:
Thanks, just merge it, i don't need the credits , as long as it works i am happy:P

PR #657, opened 2018-03-07
Clear cached options

uNknownMark:
Hello Mr. Andy. Thank you so much for your great plugin. It very helpful and great.
Looks like I found 1 place that can work incorrectly if persistent object cache used on site
Delete option in this way, will not delete them from cache. So, looks like options must be deleted via delete_site_option

### DIFF
--- a/src/GitHub_Updater/Base.php
+++ b/src/GitHub_Updater/Base.php
@@ -882,13 +882,14 @@ class Base {
 	 * @return bool
 	 */
 	public function delete_all_cached_data() {
-		global $wpdb;
+        $plugins = Singleton::get_instance( 'Plugin' )->get_plugin_configs();
+        $themes  = Singleton::get_instance( 'Theme' )->get_theme_configs();
 
-		$table         = is_multisite() ? $wpdb->base_prefix . 'sitemeta' : $wpdb->base_prefix . 'options';
-		$column        = is_multisite() ? 'meta_key' : 'option_name';
-		$delete_string = 'DELETE FROM ' . $table . ' WHERE ' . $column . ' LIKE %s LIMIT 1000';
+        $repos = array_merge( $plugins, $themes );
 
-		$wpdb->query( $wpdb->prepare( $delete_string, array( '%ghu-%' ) ) );
+        foreach( $repos as $repo ) {
+            delete_site_option( 'ghu-' . md5( $repo->repo ) );
+        }
 
 		$this->force_run_cron_job();
 


afragen:
The cache is a stored option in the database. Every option is prefixed with `ghu-`. The code as written searches the appropriate options table and deletes those entries.

Is there some place the data exists that isn’t deleted with the existing code?

Can you give an example? Screenshots of the database would be helpful.

uNknownMark:
Yes, if this cache was queried before via get_site_option, value will be putted in object cache. If object cache persistent (for example use Memcache), then value will be cached there for some time.
delete_site_option delete data from database, and also delete data from cache.
you can try enable any persistent cache for test site (for example set cache on 15 minutes), visit  github-updater page where cache will be queried, and then press clear cache. then try to get value for that deleted cache via get_site_option. And you will see that even if values not in db, but they still present.
make sense?

uNknownMark:
I created two test scripts for check this.
test1.php
`
<?php
define('WP_USE_THEMES', false);
include_once('index.php');

$test_option_name = 'sometestname';
$test_option_value = 'test_option_value';

update_site_option($test_option_name, $test_option_value);

var_dump(get_site_option( $test_option_name ));
`

test2.php

<?php
define('WP_USE_THEMES', false);
include_once('index.php');

global $wpdb;

$test_option_name = 'sometestname';

$delete_string = "DELETE FROM {$wpdb->options} WHERE option_name LIKE '%ometestna%' LIMIT 1000";
$wpdb->query( $delete_string );

var_dump(get_site_option( $test_option_name ));`
`

You can put them in root directory of your site. Visit first site.com/test1.php, then visit site.com/test2.php, and again site.com/test2.php.
Persistent object cache must be enabled. And you will that on second visit of test2.php, you will still see value for that option.

afragen:
The current code is much more efficient and I'm not sure how long that object cached data would really be present. If there is some "real world" example of stale data effecting the plugin I'll be happy to revisit this, but at this time if the object cache were to return a stale piece of data ( over 12 hours old ) the plugin would automatically recreate it with another API and get current data.

Thanks for the PR. It did cause me to go back and look at how I store the options.

https://github.com/afragen/github-updater/blob/94d03c8795bdd3471cba7d507f7a027bdd63824b/src/GitHub_Updater/API.php#L313-L325 clearly shows what happens if the returned data is expired.

uNknownMark:
At least cached data can cause problem if you push plugin new version, and then want to update it via github-updater. Instead of get update in few minutes, it will be show depend on cache expire settings.
My idea was that it can be good practice to delete options via designed for this functions (that will also care about any cache).

afragen:
I do appreciate your idea and time Mark. I’m really simply trying to optimize efficiency and honestly I have an object cache running on several sites and I don’t reall notice a problem. This seems to be a dev only issue that could be solved using the webhook updating. Take a look at the wiki to see what I mean.

uNknownMark:
Ok, thank you. At least sites that use wpengine with enabled object cache (they use memcache) can be affected

PR #661, opened 2018-03-26
Adding cache refresh to remote API

JstnPwll:
I added this because I wanted to automatically trigger a cache refresh after updating my own custom themes and plugins. On a live site it's not always advantageous to automatically install the updates after publishing...I may want to do some testing first on a non-production site before installing my latest updates. But I do want the cache refreshed so I can immediately update from my WP admin panel as soon as I feel comfortable with the new code.

### DIFF
--- a/src/GitHub_Updater/Rest_Update.php
+++ b/src/GitHub_Updater/Rest_Update.php
@@ -140,6 +140,14 @@ class Rest_Update extends Base {
 		$upgrader = new \Theme_Upgrader( $this->upgrader_skin );
 		$upgrader->upgrade( $theme->repo );
 	}
+	
+	/**
+	 * Refresh the cache.
+	 */
+	public function refresh_cache() {
+		$this->delete_all_cached_data();
+		$this->upgrader_skin->messages[] = "Cache refreshed successfully.";
+	}
 
 	/**
 	 * Is there an error?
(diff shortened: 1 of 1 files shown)


afragen:
You can already do this using WP-CLI.

`$ wp github-updater cache delete`

It sorta starts to become outside of scope for what a webhook is trying to accomplish. Convince me.

JstnPwll:
Using WP-CLI is fine if you already use WP-CLI, if not then it's nice to have an alternative means to accomplish the same thing. I personally do not want to install and use WP-CLI for one tiny task, which is why made this change for myself.

If it's not going to be helpful for any other users, feel free to close it. I'll just have to maintain my fork.

afragen:
Let me see if I can make a hook for you to do something similar that would be persistent.

afragen:
@JstnPwll use the branch switcher to re-install the `develop` branch and use the following action hook to refresh the cache.

```php
add_action( 'github_updater_pre_rest_process_request', function() {

	if ( isset( $_REQUEST['refresh_cache'] ) ) {
		$json_encode_flags = 128 | 64;

		$base = new Fragen\GitHub_Updater\Base();
		$base->delete_all_cached_data();

		header( 'HTTP/1.1 200 Expectation Failed' );
		header( 'Content-Type: application/json' );

		echo json_encode( array(
			'message'       => 'Cache refreshed successfully.',
			'cache_refresh' => true,
		), $json_encode_flags );
		exit;
	}
} );
```

Issue #663, opened 2018-03-30
GitLab Webhook Timeout Issues


Raruto:
Following the wiki at [restful-endpoints-for-remote-management](https://github.com/afragen/github-updater/wiki/Remote-Management---RESTful-Endpoints#restful-endpoints-for-remote-management) everything works as described (eg. curl related calls).
When I push (test or not) to the Gitlab repository the webhook/integration trigger is triggered I get the above mentioned error: **Hook execution failed: Net::ReadTimeout**

```
Request URL: POST http://www.domainame.com/wp-admin/admin-ajax.php?action=github-updater-update&key=76bb2b7c819c36ee37292b6978a4ad61&theme=theme-name [internal error]
Trigger: Push Hook
Elapsed time: 3.29 sec
```
instead using this testing REST endpoint
```
Request URL: POST http://webhook.site/d6eb13a3-82a5-4385-abc6-91822f508b35 [200]
Trigger: Tag Push Hook
Elapsed time: 0.35 sec
```
from the [GitLab documentation](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/doc/user/project/integrations/webhooks.md#webhook-endpoint-tips) about webhooks, I read:

> Your endpoint should send its HTTP response as fast as possible. 
> If you wait too long, GitLab may decide the hook failed and retry it.

Surfing the net i found this other [similiar issue](https://github.com/auth0-extensions/auth0-gitlab-deploy/issues/17) (which makes me think that there is a timeout after 3 seconds..)

Is it possible that this is also due to wordpress performance limits? (eg. minimum response time)
if so, it would be almost impossible to use this feature with the online version of GitLab...

Any way, thanks for the effort!

afragen:
I don’t set any timeouts so this must be happening somewhere else. 

Is the elapsed time the time it takes WP to process the webhook or the time it take GitLab to send the webhook?

Raruto:
Hi Andy,
I was referring to a connection timeout that I suppose is set on gitlab.com

As you can see gitlab requests almost always fail if it takes more than 3 seconds
![webhook-testing-example](https://user-images.githubusercontent.com/9614886/38165326-e61fe458-3511-11e8-8188-aebbc140f398.png)

Below you can see the few details available after a failed request (from gitlab.com to wordpress):

**Request details** `Net::ReadTimeout `
```
[...]

PR #666, opened 2018-04-06
Added Rest Process logging feature

Raruto:
Refers to: https://github.com/afragen/github-updater/issues/663#issuecomment-379393101

### DIFF
--- a/src/GitHub_Updater/Rest_Update.php
+++ b/src/GitHub_Updater/Rest_Update.php
@@ -160,6 +160,25 @@ class Rest_Update extends Base {
 	}
 
 	/**
+	 * Log messages during update
+	 */
+	public function log($message, $obj=""){
+
+		if(!defined('GHU_DEBUG'))
+			return;
+
+		$current_time = microtime(true);
+
+		if(!isset($this->start)) {
+			$this->start = $current_time;
+		}
+
+		$time_lapse = $current_time - $this->start;
+		$millis = round( $time_lapse * 1000 );
+		error_log( $millis . " ms : ". $message . " " . print_r( $obj, true ) );
+	}
+
+	/**
 	 * Process request.
 	 *
 	 * Relies on data in $_REQUEST, prints out json and exits.
@@ -169,6 +188,11 @@ class Rest_Update extends Base {
 	 */
 	public function process_request() {
 		try {
+
+			// DEBUG
+			$this->log("ENTERING IN \"process_request\", \$_REQUEST = ", $_REQUEST);
+			//
+
 			/*
 			 * 128 == JSON_PRETTY_PRINT
 			 * 64 == JSON_UNESCAPED_SLASHES
@@ -180,6 +204,10 @@ class Rest_Update extends Base {
 			) {
 				throw new \UnexpectedValueException( 'Bad API key.' );
 			}
+
+			// DEBUG
+			$this->log("doing action: \"github_updater_pre_rest_process_request\"");
+			//
 
 			/**
 			 * Allow access into the REST Update process.
@@ -198,6 +226,12 @@ class Rest_Update extends Base {
 
 			$current_branch   = $this->get_local_branch();
 			$webhook_response = $this->get_webhook_source();
+
+			// DEBUG
+			 $this->log("\$current_branch = ", $current_branch);
+			 $this->log("\$webhook_response = ", $webhook_response);
+			//
+
 			if ( null !== $current_branch && $tag !== $current_branch ) {
 				throw new \UnexpectedValueException( 'Request tag and webhook are not matching.' );
 			}
@@ -214,10 +248,16 @@ class Rest_Update extends Base {
 			header( 'HTTP/1.1 417 Expectation Failed' );
 			header( 'Content-Type: application/json' );
 
-			echo json_encode( array(
+			$http_response = json_encode( array(
 				'message' => $e->getMessage(),
 				'error'   => true,
 			), $json_encode_flags );
+
+			// DEBUG
+			$this->log("ENDING \"\\Exception\", \$http_response = ", $http_response);
+			//
+
+			echo $http_response;
 			exit;
 		}
 
@@ -236,7 +276,13 @@ class Rest_Update extends Base {
 			$response['success'] = true;
 		}
 
-		echo json_encode( $response, $json_encode_flags ) . "\n";
+		$http_response = json_encode( $response, $json_encode_flags ) . "\n";
+
+		// DEBUG
+		$this->log("ENDING \"process_request\", \$http_response = ", $http_response);
+		//
+
+		echo $http_response;
 		exit;
 	}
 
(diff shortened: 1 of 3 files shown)


afragen:
Thanks @Raruto I’m going to have to dig into this.

src/GitHub_Updater/Rest_Update.php
```diff
@@ -248,15 +294,17 @@ public function process_request() {
 	private function get_local_branch() {
 		if ( isset( $_REQUEST['plugin'] ) ) {
 			$repos = Singleton::get_instance( 'Plugin', $this )->get_plugin_configs();
-			$repo  = $repos[ $_REQUEST['plugin'] ];
+			$repo  = isset($repos[ $_REQUEST['plugin'] ]) ? $repos[ $_REQUEST['plugin'] ] : null;
```
afragen: This is already inside of a conditional testing this. Not sure this makes sense.

src/GitHub_Updater/Rest_Update.php
```diff
@@ -248,15 +294,17 @@ public function process_request() {
...
 		}
 		if ( isset( $_REQUEST['theme'] ) ) {
 			$repos = Singleton::get_instance( 'Theme', $this )->get_theme_configs();
-			$repo  = $repos[ $_REQUEST['theme'] ];
+			$repo  = isset($repos[ $_REQUEST['theme'] ]) ? $repos[ $_REQUEST['theme'] ] : null;
```
afragen: This is also inside a conditional of the same test.

src/GitHub_Updater/Rest_Update.php
```diff
@@ -278,5 +326,6 @@ private function get_webhook_source() {
 				break;
 		}
 		$this->upgrader_skin->messages[] = $webhook_source;
```
afragen: In most recent commit I changed this part. It’s not necessary to return the variable as it’s no longer used.

afragen:
What’s the overall benefit?

afragen:
In latest commit I have added an 'elapsed_time' to the response. While it doesn't break down the timing of each component, it is simpler. https://github.com/afragen/github-updater/commit/fdc2baa154cb9f9ac84b6fc7f5a3e53819f9bd56

Let me know what you think.

src/GitHub_Updater/Rest_Update.php
Raruto: You're definitely right, maybe due to the following: `if (isset ($ repo))` (and night developing...)

src/GitHub_Updater/Rest_Update.php
Raruto: ok, now I saw it

src/GitHub_Updater/Rest_Update.php
```diff
@@ -159,6 +159,25 @@ public function get_messages() {
 		return $this->upgrader_skin->messages;
 	}
 
+	/**
+	 * Log messages during update
+	 */
+	public function log($message, $obj=""){
+
+		if(!defined('GHU_DEBUG'))
+			return;
```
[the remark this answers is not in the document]
Raruto: at the beginning I was going to do something like that (maybe inspired by some other plugins):
```php
if(defined('GHU_DEBUG')) {
   if(!defined('WP_DEBUG')) define('WP_DEBUG',true)
   if(!defined('WP_DEBUG_LOG')) define('WP_DEBUG_LOG',true)
  //etc.
}
```
but then I let it go ..

Raruto:
> What’s the overall benefit?

My pull was intended only for:
- quickly explain the conditions in which I have carried out my tests
- recommend you to think about giving a way to debug at wordpress side

As you will have understood from what I have just said, your [commit](https://github.com/afragen/github-updater/commit/fdc2baa154cb9f9ac84b6fc7f5a3e53819f9bd56) (though it is a good idea) it assumes that the HTTP Response is correctly delivered to the Git Host

afragen:
What do you think of the latest commit? It should do everything you've outlined in this PR, I think.

afragen:
> My pull was intended only for:
> 
> quickly explain the conditions in which I have carried out my tests
> recommend you to think about giving a way to debug at wordpress side

I very good idea considering that GitLab seems to truncate the process. At least now there is some data.

src/GitHub_Updater/Rest_Update.php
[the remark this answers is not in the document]
Raruto: this [stackexchange](https://wordpress.stackexchange.com/questions/69549/define-wp-debug-conditionally-for-admins-only-log-errors-append-query-arg-f), leaving aside its question, contains some debug examples

Raruto:
What do you think of a class dedicated to debugging? (so you will not worry about what are the things tha can be "debugged"), maybe the only thing that would make this class would be just to hook certain wordpress specific hooks (or ghu specifc hooks) to debug things, for example I've read that:
- [wp_die()](https: // codex.wordpress.org/Function_Reference/wp_die) triggers a specific [hook] (https://developer.wordpress.org/?s=die&post_type%5B%5D=wp-parser-hook) before exiting, 
- and also [wp_send_json ()] (https: // codex.wordpress.org/Function_Reference/wp_send_json)

afragen:
> What do you think of a class dedicated to debugging? (so you will not worry about what are the things that can be "debugged"), maybe the only thing that would make this class would be just to "hook" certain wordpress specific hooks (or ghu specifc hooks) to debug things, for example I've read that:
> 
> wp_die() triggers a specific hook before exiting,
> and also wp_send_json()

Not sure a specific class is needed at this time. The only part of GHU that runs remotely is the webhook from a remote source. Now that data can be captured in WP. Everything else is in the browser locally for debugging and should be easily available with the proper debug settings.

src/GitHub_Updater/Rest_Update.php
Raruto: and maybe we both made a mistake...

> if ( *isset( **$_REQUEST['theme']***) ) {
> ...
> $repo  = *isset(**$repos**[ $_REQUEST['theme'] ])* 
                   ? $repos[ $_REQUEST['theme'] ]
                   : null;
> }

Raruto: Obviously it is a warning only if the theme name is wrong / non-existent

Raruto: similarly for `if (isset ($ repo))`

afragen: The latest version has improvements to where $repo is defined as false initially and if false, the $current_branch is set to 'master'.

https://github.com/afragen/github-updater/blob/e5f027b4e6a78f6ee475a60799d8308ba4d0ada5/src/GitHub_Updater/Rest_Update.php#L256-L271

Raruto: but `$repos[ $_REQUEST['theme/plugin name'] ]; ` still generates a log warning if a wrong name is supplied, am i wrong?

---
eg. Wrong theme name = "test-wrong"

[07-Apr-2018 21:33:52] PHP Notice:  Undefined index: test-wrong in "..."/plugins/github-updater/src/GitHub_Updater/Rest_Update.php on line 264

---
indeed, a you correctly pointed before, if no theme name is supplied it will never enter into the `if ( isset( $_REQUEST['theme'] ) )` condition

afragen: You are correct. https://github.com/afragen/github-updater/commit/2e1eb1328881325493bf5ab5bab6aeea66fb19d6 should fix this.

afragen:
Is there a benefit to using `wp_die`, `wp_send_json()` over what we are currently doing?

Raruto:
(stackexchange)[https://wordpress.stackexchange.com/questions/250280/best-way-to-end-wordpress-ajax-request-and-why/250282] about wp_die.
Besides, from what I understood wp_send_json like functions always end with a die() call

afragen:
https://github.com/afragen/github-updater/commit/a40ceeaf7340bb4c2cb6c7fe5ffa62a59b8c08df should take care of the remainder of this PR's concept.

PR #668, opened 2018-04-10
(Hook Logging + UI Improvements) suggestions

Raruto:
The intent of this branch is to try to increase the plugins' user friendliness (especially for newbie and non-geek users)

from https://github.com/afragen/github-updater/commit/a40ceeaf7340bb4c2cb6c7fe5ffa62a59b8c08df#commitcomment-28489018

> @Raruto I think we’ve gotten about as much simplicity and utility from webhook logging as we are able to manage, though I am always interested in improving things. I’m not sure that implementing a log system inside GHU for what happens on GitLab, GitHub, Bitbucket, etc. is a useful process. Those data are at those hosts where the webhooks are initiated. While duplicating them locally may be more convenient I’m not sure the added complexity is beneficial.

The problem, is that if I wanted to set up a system that supports Continuous Integration, I find very unpleasant "having to" keep WP_DEBUG active (just to be sure that the webhook is not causing any problems), furthermore, I think that having a double check of the functionality of a feature never hurts (ie both by Git Host and Wordpress Server)

Given the immediacy of use of the log report, I would frankly prefer to check directly on wordpress that instead every time remember (or rather forget) to go checking the webhook settings of the project of my git host (if something went wrong). 

Do you think only if I had to repeat this procedure for several different Git Hosts? 
You will say: "_simple you can use WP_DEBUG_", 
but this already implies a certain user experience and a knowledge of: php, wordpress, log files and at last of your plugin.

You will still say: "_well, but if not you have certain knowledge you do not develop a plugin/theme and then you will certainly not use GHU.._ ",
you're right, but I am convinced that certain people (especially Frontend Developers) do not really love "complicated things" (for example: dealing with the file wp-config.php and worse reading a log file..)

---

**Hook Logging Feature:** you can see most of the changes in the file: _Rest_Log_Table.php_ 
(I do not like this name for the class..)

**UI Improvements:** you can see most of the changes in the file: _Settings.php_
the most part are just links... I have not modified the .pot file because I do not know how it works, it does not seem difficult, but I did not have the time to deepen.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
 * added a `log` function into the `Rest_Update` class
 * added `GHU_DEBUG` constant support
 * fixed some undefined variable warning
+* fixed missing statment `return $webhook_source` into `get_webhook_source()`
 
 #### 7.5.0 / 2018-01-28
 * fixed _View detail_ ratings for large projects with lots of issues
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,9 @@
 * refactored code out of `class API` into specific API classes
 * simplify RESTful update code, no longer parses webhook payload just webhook itself
 * updated RESTful update code to use `site_transient_{$transient}` filter to add to update transient
+* added a `log` function into the `Rest_Update` class
+* added `GHU_DEBUG` constant support
+* fixed some undefined variable warning
 
 #### 7.5.0 / 2018-01-28
 * fixed _View detail_ ratings for large projects with lots of issues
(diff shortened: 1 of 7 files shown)


Raruto:
Llooking at it quickly, I realized that:
- I forgot to delete this site option "**ghu_db_version**" when the plugin is deleted (in theory it should remain saved from one update to the next)
- you might argue that these calls `Rest_Log_Table::insert_db_record()` and `error_log()` offset the calculation of the time lapse, but it is enough to understand what time we want to estimate (in this way it is nothing more than the calculation of the duration of the function `process_request()`)

Raruto:
based on version 7.6.0 (I think), I also wrote the class diagram:
## Brief
class diagram - brief
## Verbose
classdiagram - verbose

---

I do not think it's really "complete" (for example the _Singleton_ class is missing), but I hope somehow it can helps you, for example:
- it can be used as a starting point to write tests
- it can helps you decide if you're going to make some code refactoring
- it can helps other developers dig faster in your project
- or maybe in your next wordcamp...

if you're interested in changing it, I can also commit the uml project file (I have used [StarUML](http://staruml.io/))

---

Speaking of _Singleton_ class, can I ask you if the pattern you used has a specific name? and is there a specific reason why you use it?

---

**PS** take all the time necessary to answer, I have no urgency.. 😷

Issue #672, opened 2018-04-19
Gitlab CE/Enterprise Access Token Alert Persists


evhaddad:
When configuring the plugin, I am prompted to enter a GitLab.com, GitLab CE, or GitLab Enterprise Access Token. However even after entering a Gitlab CE Access Token, the alert does not go away. Is this expected behavior?

<img width="1426" alt="screen shot 2018-04-19 at 12 07 14 pm" src="https://user-images.githubusercontent.com/38531692/39016641-05ec67b6-43ef-11e8-9221-adeadfa238a5.png">

afragen:
Thanks Elizabeth. I’ll look into it.

afragen:
Just making notes as my laptop is in for repairs. ☹️ 

Adjust `Base::get_running_git_servers()` to differentiate GitLab.com and GitLab self-hosted. Then adjust `GitLab::set_default_credentials()`

afragen:
Thanks @pborreli. I do have an idea on how to fix this. It’s just that my laptop is out for repairs at the moment. 😉

PR #673, opened 2018-04-20
Proposal for Fixing issue #672

pborreli:
Just tested with my own test case : self hosted gitlab

### DIFF
--- a/src/GitHub_Updater/API/GitLab_API.php
+++ b/src/GitHub_Updater/API/GitLab_API.php
@@ -73,7 +73,8 @@ class GitLab_API extends API implements API_Interface {
 			$set_credentials                            = true;
 		}
 		if ( ( empty( static::$options['gitlab_enterprise_token'] ) &&
-		       ! empty( $this->type->enterprise ) ) ||
+		       ! empty( $this->type->enterprise ) &&
+		       'gitlab.com' === $this->type->enterprise ) ||
 		     ( empty( static::$options['gitlab_access_token'] ) &&
 		       in_array( 'gitlab', $running_servers, true ) )
 		) {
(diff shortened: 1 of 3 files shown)


src/GitHub_Updater/API/GitLab_API.php
```diff
@@ -73,7 +73,8 @@ protected function set_default_credentials() {
 			$set_credentials                            = true;
 		}
 		if ( ( empty( static::$options['gitlab_enterprise_token'] ) &&
-		       ! empty( $this->type->enterprise ) ) ||
+		       ! empty( $this->type->enterprise ) &&
```
afragen: This line would be `in_array( ‘gitlab_enterprise’, $running_servers, true ) &&`

afragen: Assuming the change below.

src/GitHub_Updater/Settings.php
```diff
@@ -134,6 +134,8 @@ private function settings_sub_tabs() {
 		}
 
 		foreach ( $gits as $git ) {
+			$parts = explode('-', $git);
+			$git = $parts[0];
```
afragen: Do we really need this? Shouldn’t the “GitLab” tab already display?

src/GitHub_Updater/Base.php
```diff
@@ -1270,6 +1270,9 @@ public function get_running_git_servers() {
 			if ( ! empty( $e->enterprise ) && false !== stripos( $e->type, 'bitbucket' ) ) {
 				return 'bbserver';
 			}
+			if (! empty( $e->enterprise ) && 'gitlab.com' !== $e->enterprise ) {
+				return 'gitlab-selfhosted';
+			}
```
afragen: Let’s stick to the same logic for the conditional using `stripos( $e->type, ‘gitlab’ )` and return `gitlab_enterprise` as above.

pborreli:
not convinced at all by this solution, made more test and it doesn't work when more there is more than 1 repo on gitlab hosted

PR #680, opened 2018-05-12
Fixes CLI_Installer_Skin not found when calling Install::install( $type, $array ) outside from wp-cli

Raruto:
**Short description:**

calling the following function outside of the WP-CLI's context:
```
$installer = Singleton::get_instance( 'Fragen\GitHub_Updater\Install', $this );
$installer->install( 'plugin', array( 'uri' => 'https://github.com/afragen/test-plugin' ) );
```
throws:
```
Uncaught Error: 
Class 'Fragen\GitHub_Updater\CLI_Plugin_Installer_Skin' not found 
in wp-content\plugins\github-updater\src\GitHub_Updater\Install.php:183
```
this happens because the execution flow depends on whether the Install::install( $type, $array ) function is called from WP-CLI or not:

to deepen see the following lines:
- https://github.com/afragen/github-updater/blob/318f09f0a7e1a08774dafb46cf8289968bddaaba/src/GitHub_Updater/Install.php#L66
- https://github.com/afragen/github-updater/blob/318f09f0a7e1a08774dafb46cf8289968bddaaba/src/GitHub_Updater/Install.php#L183
- https://github.com/afragen/github-updater/blob/318f09f0a7e1a08774dafb46cf8289968bddaaba/src/GitHub_Updater/Install.php#L199
- https://github.com/afragen/github-updater/blob/318f09f0a7e1a08774dafb46cf8289968bddaaba/src/GitHub_Updater/Init.php#L32

---

this simple fix adds a third parameter to the install function, modifying it as follows: `install ( $type, $config = null, $is_wp_cli = false )`

anyway, some kind of refactoring to the install function should be done ( it's really a big function.. )


afragen:
OK a little refactoring completed. https://github.com/afragen/github-updater/commit/65602dc92cd2f71a1a4e40b3a0b0c9fc7cb5ab5b

Raruto:
> With the latest commits on my end there should no longer be an error for the skin not being found.

No, because it requires that the wp-cli stuff is included to work properly (obviously calling the function out of the wp-cli's context, ie not in a terminal..)

https://github.com/afragen/github-updater/blob/318f09f0a7e1a08774dafb46cf8289968bddaaba/src/GitHub_Updater/Init.php#L32
https://github.com/afragen/github-updater/blob/6e63cfbd88c10d632f013fa3eb4c55f8009a5295/src/GitHub_Updater/Traits/GHU_Trait.php#L36

---

[I slightly modified](https://github.com/afragen/github-updater/pull/680/commits/d34067ed3ca2a0d385edf97d0a262d39ec94d0e1) your last commit:
```php
private function set_wpcli_install_post_data( $wp_cli_config ) {
	if ( ! static::is_wp_cli() ) { // <-- this condition is too strong for what currently does this function
		return;
	}
	...
```
 so I can call:
`$ installer-> install ('plugin', array ('uri' => 'https://github.com/afragen/test-plugin'));`
without throwing errors

Raruto:
> There are 2 methods within GHU to install a plugin/theme.

I know, but:

>1. From the Install Plugin/Theme tab in the Settings page.

it requires **$_POST manipulation** (which is however something that this function should not take care of..)

>2. From WP-CLI

it requires **wp-cli installed** (and included..)

---

>In what setting are you needing to call 
>`$installer->install( 'plugin', array( 'uri' => 'https://github.com/afragen/test-plugin' ) );`?

I simply want to call this function outside of GHU's scope.

```php
namespace myNamespace;

class MyClass {
  ...
  public function myFunc( $array ) {
    $installer = Singleton::get_instance( 'Fragen\GitHub_Updater\Install', $this );
    $installer->install( 'plugin', $array );
  }
  ...
}

$instance = new MyClass( ... );
$instance->myFunc( array( 'uri' => 'https://github.com/afragen/test-plugin' ) );
```
in [this way](https://github.com/afragen/github-updater/pull/680/commits/d34067ed3ca2a0d385edf97d0a262d39ec94d0e1) we have a stateless call (eg without **$_POST** manipulation) and we can use it in plugins without "strange" calls..

---

Anyway, this:
https://github.com/afragen/github-updater/blob/65602dc92cd2f71a1a4e40b3a0b0c9fc7cb5ab5b/src/GitHub_Updater/Install.php#L216
should be moved into this function:
https://github.com/afragen/github-updater/blob/65602dc92cd2f71a1a4e40b3a0b0c9fc7cb5ab5b/src/GitHub_Updater/Install.php#L106

afragen:
Help me understand the use case. How would be used in another plugin?

Raruto:
In this way, WP-CLI and PHP calls share the same commands/parameters:
```php
 * <uri>
 * : URI to the repo being installed
 *
 * [--branch=<branch_name>]
 * : String indicating the branch name to be installed
 * ---
 * default: master
 * ---
 *
 * [--token=<access_token>]
 * : GitHub, GitLab, or Gitea access token if not already saved
 *
 * [--bitbucket-private]
 * : Indicates a private Bitbucket repository
 *
 * [--github]
 * : Optional to denote a GitHub repository
 * Required when installing from a self-hosted GitHub installation
 *
 * [--bitbucket]
 * : Optional switch to denote a Bitbucket repository
 * Required when installing from a self-hosted Bitbucket installation
 *
 * [--gitlab]
 * : Optional switch to denote a GitLab repository
 * Required when installing from a self-hosted GitLab installation
 *
 * [--gitea]
 * : Optional switch to denote a Gitea repository
 * Required when installing from a Gitea installation
```
for example: 
```
 wp plugin install-git https://github.com/afragen/test-plugin --branch=develop --token=...
```

```php
$array = array( 
  'uri'    => 'https://github.com/afragen/test-plugin', 
  'token'  => '...', 
  'branch' => 'develop',
  //etc.
);
Singleton::get_instance( 'Fragen\GitHub_Updater\Install', $this )->install( 'plugin', $array );
```

Before you tell me, I already know that I could do similar things with [wp_dependency_installer](https://github.com/afragen/wp-dependency-installer), but:
- as you said: "it only works with public repositories".
- as you said: "you like little and independent plugins".
- as i said: "i don't like to depend on too many plugins".
- as i said: "I'd like to minimize future efforts..."
- as you said: "write your own plugin so you can easily maintain it..".

---

At the moment, I'm just trying to implement the mass import/export feature (based on my testing branch) of which I spoke earlier..

---

anyway, what sense would have added the opportunities to add own tabs/subtabs to GHU without having a simple manner to call the GHU's API? (eg stateless function calls)

afragen:
OK, you convinced me. 👍

PR #721, opened 2018-10-04
Do not prepend base_uri to full GitHub Enterprise URLs

rlindner:
When providing a full URL to an GitHub Enterprise hosted repo, e.g. `https://:GHE_HOST/api/v3/repos/:GHE_ORG/:GHE_REPO`, the used URL resolves to an invalid URL `https://api.github.comhttps://:GHE_HOST/api/v3/repos/:GHE_ORG/:GHE_REPO/contents/style.css`, because the API host is always being prepended.


afragen:
Thanks! I still need to test but don’t see any reason not to merge.

PR #726, opened 2018-10-30
Fix for 1password on Chrome autofilling the GitHub Access Token field

figureone:
The 1Password browser extension on Chrome tries to be too smart, and autofills any `GitHub Access Token` fields with the credentials for the WordPress site itself:

image

This pull request simply uses the suggested practice from MDN to prevent autocomplete on password fields that aren't meant to be autofilled.
https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#The_autocomplete_attribute_and_login_fields

Thanks!

### DIFF
--- a/src/GitHub_Updater/API/Bitbucket_API.php
+++ b/src/GitHub_Updater/API/Bitbucket_API.php
@@ -491,7 +491,7 @@ class Bitbucket_API extends API implements API_Interface {
 	public function bitbucket_password() {
 		?>
 		<label for="bitbucket_password">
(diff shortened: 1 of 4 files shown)


afragen:
Interesting. Can you update the PR for GitLab and Gitea classes?

figureone:
Sure! Extended to Bitbucket, GitLab, and Gitea.

afragen:
Mahalo

PR #760, opened 2019-01-23
add a quick link in the admin bar to the refresh transients feature

squarecandy:
I find I'm constantly needing to access this feature on sites I manage with GitHub Updater. It would be convenient to be able to access this directly from any place in WordPress - front or back end. This would turn a process of 2 or 3 clicks/page loads into 1 click.

The only thing I'm not sure of it compatibility w/ multisite, which I see is important based on your other code. I don't think `network_admin_bar_menu` exists, does it?

### DIFF
--- a/src/GitHub_Updater/Settings.php
+++ b/src/GitHub_Updater/Settings.php
@@ -88,6 +88,7 @@ class Settings extends Base {
 	 */
 	protected function load_hooks() {
 		add_action( is_multisite() ? 'network_admin_menu' : 'admin_menu', [ $this, 'add_plugin_page' ] );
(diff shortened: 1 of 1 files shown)


afragen:
@squarecandy I'm not sure this is needed in the main plugin. It seems like a great idea for an add-on plugin.

I can definitely link to it from the [Extras and Credits wiki page](https://github.com/afragen/github-updater/wiki/Extras-and-Credits).

squarecandy:
That makes sense - I'll make a separate plugin for it.

squarecandy:
I think I don't need the button since you showed me how to setup webhooks to push it to the sites automatically... Thanks for your help today.

afragen:
Anytime.

PR #772, opened 2019-03-07
Enhance matching for repo slugs

sc0ttkclark:
Allow for soft or exact matches but don't take the first rename match as the final match. Also allow for plugin directory to be used as match, not just `$repo->slug`


sc0ttkclark:
Just opcode optimizations:

https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

PhpStorm and others make recommendations to use it as well. It's only useful when using namespaces.

afragen:
Hmm, it would be really nice if PHPCS picked this up.

PR #807, opened 2019-09-23
Repos with a '.' in their name were not re-connected properly

ymauray:
I have a repo on github that has a '.' in its name. On GHU's settings page, I had this error message : 
image

This simple fix uses `basename` - which contains the complete file name, with extension - rather than `filename` - which does not contains the extension.

### DIFF
--- a/src/GitHub_Updater/Traits/GHU_Trait.php
+++ b/src/GitHub_Updater/Traits/GHU_Trait.php
@@ -331,7 +331,7 @@ trait GHU_Trait {
 		$header['scheme']     = isset( $header_parts['scheme'] ) ? $header_parts['scheme'] : null;
 		$header['host']       = isset( $header_parts['host'] ) ? $header_parts['host'] : null;
 		$header['owner']      = trim( $header_path['dirname'], '/' );
-		$header['repo']       = $header_path['filename'];
+		$header['repo']       = $header_path['basename'];
 		$header['owner_repo'] = implode( '/', [ $header['owner'], $header['repo'] ] );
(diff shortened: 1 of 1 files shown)


afragen:
The problem here is that many people add use something like `afragen/my-git-repo.git` in their `GitHub Plugin URI`. 

The suggested fix is to not use dots in your repo name as they can be misinterpreted. Unfortunately your patch does not solve this issue.

afragen:
I just had a thought on how to possible fix this. Maybe something like the following.

`$header['repo'] = 'git' === $header_path['extension'] ? $header_path['filename'] : $header_path['basename'];`

I’ll have to test this extensively as it can seriously break things.

ymauray:
Ah, interesting. I didn't think about the ".git" extension being used. 
How about an config option then ?
That way the default behaviour is not changed. And it can be enabled on a per repo basis.

afragen:
I really don’t wish to confuse people who innocently add the `.git` to their URI with some new option. Unfortunately it’s not an insignificant number. I would much rather have something in code that solves the issue.

Thus far my response has been to change the repo name to use a `-` instead of a `.`

ymauray:
Fair enough. I'll keep that in mind next time I create a repo !

afragen:
> Fair enough. I'll keep that in mind next time I create a repo !

If we can make this work it shouldn’t matter. 😉

afragen:
@ymauray you can test the `develop` branch and see if that works for you. It seems to work fine in my testing.

ymauray:
Yes, it looks ok on my end too !

PR #829, opened 2019-12-16
Avoid unnecessary function calls within a for-loop

chesio:
Function calls are quite expensive, so when their return values do not depend on iterated item, they can be computed only once to spare some CPU time.

Use of variables in this case makes the if conditions more readable (but that's just my opinion).

### DIFF
--- a/src/GitHub_Updater/Plugin.php
+++ b/src/GitHub_Updater/Plugin.php
@@ -226,6 +226,10 @@ class Plugin {
 	 * Calls to remote APIs to get data.
 	 */
 	public function get_remote_plugin_meta() {
+		$can_add_plugin_row_action = ( 'init' === current_filter() ) && ( ! is_multisite() || is_network_admin() );
+		$github_updater_disable_wpcron = apply_filters( 'github_updater_disable_wpcron', false );
+		$is_wp_cli = static::is_wp_cli();
+
 		$plugins = array();
 		foreach ( (array) $this->config as $plugin ) {
 			/**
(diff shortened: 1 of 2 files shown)


src/GitHub_Updater/Plugin.php
```diff
@@ -257,13 +257,13 @@ public function get_remote_plugin_meta() {
 		if ( $schedule_event && ! empty( $plugins ) ) {
 			if ( ! wp_next_scheduled( 'ghu_get_remote_plugin' ) &&
 			! $this->is_duplicate_wp_cron_event( 'ghu_get_remote_plugin' ) &&
-			! apply_filters( 'github_updater_disable_wpcron', false )
+			! $github_updater_disable_wpcron
```
afragen: This isn't in the foreach loop.

src/GitHub_Updater/Plugin.php
```diff
@@ -257,13 +257,13 @@ public function get_remote_plugin_meta() {
...
 			) {
 				wp_schedule_single_event( time(), 'ghu_get_remote_plugin', array( $plugins ) );
 			}
 		}
 
-		if ( ! static::is_wp_cli() ) {
+		if ( ! $is_wp_cli ) {
```
afragen: This isn't in the foreach loop.

afragen:
I understand what the intent is here, but does the abstraction actually provide a benefit? Having the actual functions in those locations does provide for ease of reading the code.

What is the actual speed benefit? Are a a few CPU cycles worth it here?

afragen:
So I've tested this commit and the times are as follows.

The total time for all the functions, run once, ranges run from 1.68E-5 to 2.38E-5 microseconds. I'm not certain savings is worth it. 

Running these 1000x is only 0.17 microseconds, 100x = 0.0018 microseconds. 

I seriously doubt anyone will have the need to loop that many times.

chesio:
> I seriously doubt anyone will have the need to loop that many times.

Me either. The point is more about general good practice to keep loop-invariant calculations outside of the loop. I haven't run into any performance issues with the plugin that would make me to hunt for optimizations, it's just something that always strikes me when I see it while scanning a codebase :-)

Another reasoning is that when using filters you have no control over what someone hooks to `github_updater_disable_wpcron` filter - surely it will be `__return_true` method most of the time, but it can be a method that reads from database for some reason, so I would avoid calling `apply_filters('XYZ', ...)` multiple times in the same method. But again, it's more about general good practice than this particular case.

As you find existing code more readable, I'm closing the PR.

PR #864, opened 2020-03-09
Make sure options indexes are set before accessing them.

te-online:
I saw this notice in my wp-debug-log:

`PHP Notice: Undefined index: github_access_token in wordpress/wp-content/plugins/github-updater/src/GitHub_Updater/Traits/Basic_Auth_Loader.php on line 145`

And I believe this change will fix that.

### DIFF
--- a/src/GitHub_Updater/Traits/Basic_Auth_Loader.php
+++ b/src/GitHub_Updater/Traits/Basic_Auth_Loader.php
@@ -152,20 +152,20 @@ trait Basic_Auth_Loader {
 				break;
 			case 'github':
 			case $type instanceof GitHub_API:
-				$token = ! empty( $options['github_access_token'] ) ? $options['github_access_token'] : null;
-				$token = ! empty( $options[ $slug ] ) ? $options[ $slug ] : $token;
+				$token = isset( $options['github_access_token'] ) && ! empty( $options['github_access_token'] ) ? $options['github_access_token'] : null;
+				$token = isset( $options[ $slug ] ) && ! empty( $options[ $slug ] ) ? $options[ $slug ] : $token;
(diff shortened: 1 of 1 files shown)


afragen:
@te-online can you describe the steps required to reproduce these PHP notices?

te-online:
Well, I just installed the plugin, honestly. 
Then I installed one public plugin from Github https://github.com/te-online/wordpress-plugin--themeable-calendar.
After that I saw the notice a couple of times on almost every page-load (or more probably on every cron-run).

afragen:
I'll see if I can reproduce this.

te-online:
Thanks you! 😊 

It's version 9.3.1, though, that the site is using. No update is shown, even after clearing the cache. Although I can see a release for 9.3.2 here. And it's running on WordPress 5.3.2.

afragen:
BTW, your plugin won't update correctly as you have changed your tagging format. The `v` in the version number doesn't sort higher than a number when using `array_sort`. It would be better to stick to semantic versioning.

I tried it on an existing site and I got no errors. I'll need to try it on a new install.

afragen:
This was fixed in https://github.com/afragen/github-updater/commit/7c3bf47264ff28bc29777e7f6b9f709c1a132ad5

te-online:
> BTW, your plugin won't update correctly as you have changed your tagging format. The v in the version number doesn't sort higher than a number when using array_sort. It would be better to stick to semantic versioning.

Thanks! I noticed that, but didn't know what the reason was. I went for using a branch head instead of the version, but I'll change the versioning in the next release.

> This was fixed in 7c3bf47

Great stuff! Sorry for patching a fix onto an existing fix 😉 
Glad this is sorted.

PR #865, opened 2020-03-12
Fix/missing theme style.css

### LINKED ISSUE #862: CLI Declaration PHP Warning
When running latest Local Lightning (Local Lightning v5.2.5+2498, nginx, PHP 7.4.1, MySQL 8.0.16) I see this WP_DEBUG_LOG message:

```
PHP Warning:  Declaration of Fragen\GitHub_Updater\WP_CLI\CLI_Theme_Installer_Skin::feedback($string) should be compatible with WP_Upgrader_Skin::feedback($string, ...$args) in /wp-content/plugins/github-updater/src/GitHub_Updater/WP_CLI/CLI_Integration.php on line 298
PHP Stack trace:
PHP   1. {main}() /Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar:0
PHP   2. include() /Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar:4
PHP   3. include() phar:///Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar/php/boot-phar.php:11
PHP   4. WP_CLI\bootstrap() phar:///Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar/vendor/wp-cli/wp-cli/php/wp-cli.php:27
PHP   5. WP_CLI\Bootstrap\LaunchRunner->process() phar:///Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar/vendor/wp-cli/wp-cli/php/bootstrap.php:74
PHP   6. WP_CLI\Runner->start() phar:///Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar/vendor/wp-cli/wp-cli/php/WP_CLI/Bootstrap/LaunchRunner.php:23
PHP   7. WP_CLI\Runner->load_wordpress() phar:///Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php:1158
PHP   8. require() phar:///Applications/Local.app/Contents/Resources/extraResources/bin/wp-cli/wp-cli.phar/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php:1237
PHP   9. include_once() /wp-settings.php:360
PHP  10. require_once() /wp-content/plugins/github-updater/github-updater.php:53
PHP  11. Fragen\GitHub_Updater\Bootstrap->run() /wp-content/plugins/github-updater/src/GitHub_Updater/Bootstrap.php:20
PHP  12. Fragen\GitHub_Updater\Init->run() /wp-content/plugins/github-updater/src/GitHub_Updater/Bootstrap.php:58
PHP  13. include_once() /wp-content/plugins/github-updater/src/GitHub_Updater/Init.php:73
```

cliffordp:
Fixed #862 for me

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 * fix potential PHP warning in `Basic_Auth_Loader::get_slug_for_credentials()` when installer, like TGMPA, passes as array and not string
 * define `$error_code[{git}]['git']` for certain errors to avoid PHP undefined index warning
 * update calls for Bitbucket Server REST API v7, thanks @Idealien
+* ignore themes without a root `style.css` file (avoids PHP warnings), thanks @cliffordp
 
 #### 9.3.2 / 2020-02-19
 * fixed some PHP warnings and 401 errors when access tokens not set in `Basic_Auth_Loader`
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 * fix potential PHP warning in `Basic_Auth_Loader::get_slug_for_credentials()` when installer, like TGMPA, passes as array and not string
 * define `$error_code[{git}]['git']` for certain errors to avoid PHP undefined index warning
 * update calls for Bitbucket Server REST API v7, thanks @Idealien
-* ignore themes without a root `style.css` file (avoids PHP warnings), thanks @cliffordp
+* explicitly ignore themes without a root `style.css` file to avoid PHP warnings, thanks @cliffordp
 
 #### 9.3.2 / 2020-02-19
 * fixed some PHP warnings and 401 errors when access tokens not set in `Basic_Auth_Loader`
--- a/src/GitHub_Updater/Theme.php
+++ b/src/GitHub_Updater/Theme.php
@@ -254,8 +254,8 @@ class Theme {
 
 		if ( $schedule_event && ! empty( $themes ) ) {
 			if ( ! wp_next_scheduled( 'ghu_get_remote_theme' ) &&
-			! $this->is_duplicate_wp_cron_event( 'ghu_get_remote_theme' ) &&
-			! apply_filters( 'github_updater_disable_wpcron', false )
+				! $this->is_duplicate_wp_cron_event( 'ghu_get_remote_theme' ) &&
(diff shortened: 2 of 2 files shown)


afragen:
According to the theme guidelines the minimum requirements for and theme are an index.php and a styles.css. These exist in the root folder. 

If you’re using some sort of build process then I would suggest a _placeholder_ style.css that keeps the version number current and uses a release asset.

cliffordp:
I don't disagree, but GitHub Updater blows up my site if this is the case. It's simple code to avoid many notices on every single page. As the changelog and PHP text says, these directories found in the _themes_ folder will be ignored by GitHub Updater, which is in line with your pointing to WP's guidelines that it should have a style.css

afragen:
I curious. How does this theme work if there's no style.css in the root?

afragen:
If the goal is to have GHU skip this theme, why not just remove the `GitHub Theme URI` header?

cliffordp:
It isn't a valid theme folder and there isn't any github theme uri in the header (there is no header). And it didn't used to throw these errors so something changed since whatever that other version was that I had that wasn't updating itself.

So if you just create a `wp-content/themes/dummy` folder (I assume no files need to be within in), it should cause the errors.

afragen:
I'll see if I can reproduce. Can I ask why you have a non-functional theme in the theme folder?

cliffordp:
It's for template overrides -- so a folder to drop into my child theme -- that I want to test with and without (test my changes). I actually saw other folders in the array when going through this code but didn't pay attention to which they were or why they were in there.

The point of this is that even if I'm doing something dumb or temporary, the site shouldn't blow up.

cliffordp:
thanks! :)

You've been great. Thanks for your time and effort.

afragen:
Anytime @cliffordp, always happy to work with the great folks from @moderntribe

cliffordp:
We ❤️ you, too! ;)

Thanks for all your open source contributions.

PR #869, opened 2020-04-05
Fix arrow function syntax errors on IE11

sharevb:
The arrow function syntax is not supported on IE11 browser and cause JS errors.
IE11 should no longer be used but these errors have been logged on some website of Community Supported Agriculture sites where my own plugin is used.
The only subtitle difference between function and arrow function is the inclusion of this context.


afragen:
Thanks I’ll take a look at this soon.

afragen:
Thanks @sharevb!

PR #918, opened 2020-12-09
Add release workflow

shivammathur:
This PR adds a release workflow which will run when a new tag is pushed. 
It will create a release with the same name as the tag and attach the asset.


afragen:
You are awesome. Thank you so much.

PR #940, opened 2021-04-11
Polyfill for NodeList.forEach

sharevb:
Polyfill for NodeList.forEach browser that supports ES5 like older IE

### DIFF
--- a/js/ghu-install-vanilla.js
+++ b/js/ghu-install-vanilla.js
@@ -19,6 +19,21 @@
 		}
 	    };
 	}
+	
+	//polyfill for Element.matches and Element.closest in IE
+	if (!Element.prototype.matches)
+	    Element.prototype.matches = Element.prototype.msMatchesSelector ||
+					Element.prototype.webkitMatchesSelector;
+	if (!Element.prototype.closest)
+	    Element.prototype.closest = function(s) {
+		var el = this;
+		if (!document.documentElement.contains(el)) return null;
+		do {
+		    if (el.matches(s)) return el;
+		    el = el.parentElement || el.parentNode;
+		} while (el !== null && el.nodeType == 1);
+		return null;
+	    };
 
 	// Hide non-default (Bitbucket & GitLab) settings on page load.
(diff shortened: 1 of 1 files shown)


afragen:
I confess to not being proficient in javascript. Can you explain the purpose and function of this PR?

Thanks.

sharevb:
Some old browsers like IE11 don't have an implementation of `forEach` on `NodeList` and thus there is an JS error at [js/ghu-install-vanilla.js#L81](https://github.com/afragen/git-updater/blob/5b70fddaa00b4acbd8a00c40804d13ea741d5223/js/ghu-install-vanilla.js#L81). These polyfills implements `forEach` and `closest` of `Element` (in `element = item.parentElement.closest(selector);` above) for browser that don't have already.

I admit that IE11 must no longer be used but I logged such a JS error.

afragen:
There is a plan to drop support for IE11 within WordPress entirely. https://make.wordpress.org/core/2021/03/04/discussion-dropping-support-for-ie11/

afragen:
Is there a way to add this polyfill as a separate file?

sharevb:
> There is a plan to drop support for IE11 within WordPress entirely. https://make.wordpress.org/core/2021/03/04/discussion-dropping-support-for-ie11/

Yes, I agree totally that IE11 should not be used anymore

> Is there a way to add this polyfill as a separate file?

May be but in fact, those polyfill "fills gaps" only if there is not an implementation of the given method and do nothing on other browser. So I don't think it is necessary to add one more JS for this.

afragen:
@sharevb I've made a very large change to the `develop` branch. Much of it revolving around a rebrand. As such the file name for the JS file has changed. Can you pull the latest develop branch and adjust your PR?

If it's too much trouble, let me know and I'll pull it in.

sharevb:
> @sharevb I've made a very large change to the `develop` branch. Much of it revolving around a rebrand. As such the file name for the JS file has changed. Can you pull the latest develop branch and adjust your PR?
> 
> If it's too much trouble, let me know and I'll pull it in.

Yes, no problem, done in #942

PR #942, opened 2021-04-17
Polyfill for IE

sharevb:
Polyfill for NodeList.forEach, Element.matches and Element.closest in IE


sharevb:
Redone of #940 to match new project structure

afragen:
Thanks 🤩

PR #985, opened 2022-03-09
Fix plugins icons in PNG file format

### LINKED ISSUE #984: Plugin icons not behaving as expected
If a plugin hosted on GitHub has `/assets/` `icon-[128x128|256x256].jpg` or `icon.svg` all works as expected.

But for `icon-[128x128|256x256].png` it's not.

The [foreach](https://github.com/afragen/git-updater/blob/develop/src/Git_Updater/Plugin.php#L218) always checks for `.jpg` last, and sets to null if none found, even if there was a `.png` found in the previous loop.

The solution is to only set the array key if the file is found.

@afragen does it need an array key with `null` value if no icon is found?

pedro-mendonca:
Only add plugin icon key if file exists.
This avoid overriding an existing a .png icon with null from a non-existing .jpg icon.

Fixes #984

### DIFF
--- a/src/Git_Updater/Plugin.php
+++ b/src/Git_Updater/Plugin.php
@@ -216,10 +216,10 @@ class Plugin {
 				'2x_jpg' => 'icon-256x256.jpg',
 			];
 			foreach ( $icons as $key => $filename ) {
-				$key                         = preg_replace( '/_png|_jpg/', '', $key );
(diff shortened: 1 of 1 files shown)


afragen:
Fixed in 45ef89735605755c74cd91db1a2a29c522e3db43

PR #986, opened 2022-03-10
noopenernoreferer won't work

Presskopp:
maybe this is not even needed


afragen:
I think this function is a holdover. From how long ago I don't remember, but it doesn't seem to be called from anywhere.

Thanks for bringing it to my attention. I'll just remove the function. 😉

afragen:
My mistake. I found it.

Will merge and thanks.

PR #1048, opened 2023-09-09
Add title (hover) text to `plugins.php` icon, "Updates via Git Updater"

BrianHenryIE:
Add "Updates via Git Updater" to the GitHub icon on `plugins.php`.

----

Screenshot 2023-09-08 at 7 34 46 PM

----

So, I have a plugin which also uses the GitHub icon on `plugins.php` and I thought my site had gone mad because I removed my plugin but the GitHub icon persisted! I searched and searched and considered asking for help, until I eventually realised that your plugin was adding the same icon! 

https://github.com/BrianHenryIE/bh-wp-plugins-page


afragen:
Thanks Brian

PR #1049, opened 2023-09-10
Suggestion: Move the plugins.php GitHub icon to the auto-updates column

BrianHenryIE:
Change:

Screenshot 2023-09-10 at 3 44 27 PM

to:

Screenshot 2023-09-10 at 3 50 36 PM

The icon is currently added to the `plugin_row_meta` list. This instead appends it to the `plugin_auto_update_setting_html` (there is no filter for the array of items).

Personally, I consider the first column (actions) to be links internal to the site, the second (meta) to be external links, and since the relatively new third column's topic is automatic updates, it seems appropriate for an icon representing an updates function.

### DIFF
--- a/src/Git_Updater/Base.php
+++ b/src/Git_Updater/Base.php
@@ -621,6 +621,8 @@ class Base {
 	/**
 	 * Add git host based icons.
 	 *
+	 * @hooked theme_row_meta
+	 *
 	 * @param array  $links Row meta action links.
 	 * @param string $file  Plugin or theme file.
 	 *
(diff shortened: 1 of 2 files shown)


afragen:
I like it where it is.

afragen:
It's not so much that the icon represents where the update comes from. It represents where the repo is hosted. 

There are icons for Bitbucket, GitLab, and Gitea.

BrianHenryIE:
I add links to GitHub on my plugins, like:

Screenshot 2023-09-10 at 4 22 24 PM

So in this case, the icon Git Updater adds (not present in screenshot which was quickly grabbed from a local wp-env) is representative of where the update is coming from, slightly distinct from "where the repo is hosted". While it is a nice touch to communicate where the repo is hosted, the business of the plugin is updating, hence I think the Automatic Updates column is most appropriate.

But your preference is fine. I figured it was a small enough suggested change that I should open a PR rather than an issue when bringing it up.

I'd prefer if that column were called just Updates, not Automatic Updates. I've been meaning to add update install date tracking to the [Plugins Page](https://github.com/BrianHenryIE/bh-wp-plugins-page) plugin I mentioned yesterday and I'll add its UI in that column.

Thanks for Git Updater, it's a nice piece of work. I also just installed oEmbed Gists and Files yesterday and am subscribed to the wp-dependency-installer Trac ticket. Appreciate your work!

PR #1160, opened 2026-06-02
Fix: use esc_url() instead of esc_attr() on URL values in Settings.php

thisismyurl:
Hey Andy,

Three URL-context attributes in the settings page template are using `esc_attr()` where `esc_url()` is the correct function:

1. `img src` — `$logo` from `plugins_url()`
2. First `form action` — `$action` (relative admin URL)
3. Second `form action` — `$refresh_transients` from `add_query_arg()`

`esc_attr()` handles HTML attribute encoding but doesn't validate or normalise the URL. `esc_url()` does both. WordPress Plugin Check flags this pattern; it's a three-character change on each line.

(full disclosure, I used AI help with the testing and tweaks - Christopher)

### DIFF
--- a/src/Git_Updater/Settings.php
+++ b/src/Git_Updater/Settings.php
@@ -305,7 +305,7 @@ class Settings {
 			<?php $this->admin_page_notices(); ?>
 			<?php if ( 'git_updater_settings' === $tab ) : ?>
 				<?php $this->options_sub_tabs(); ?>
-				<form class="settings" method="post" action="<?php echo esc_url( $action ); ?>">
+				<form class="settings" method="post" action="<?php echo esc_attr( $action ); ?>">
 					<?php
 					settings_fields( 'git_updater' );
 					if ( 'git_updater' === $subtab ) {
--- a/src/Git_Updater/Settings.php
+++ b/src/Git_Updater/Settings.php
@@ -320,7 +320,7 @@ class Settings {
 					?>
 				</form>
 				<?php $refresh_transients = add_query_arg( [ 'git_updater_refresh_transients' => true ], $action ); ?>
-				<form class="settings" method="post" action="<?php echo esc_url( $refresh_transients ); ?>">
+				<form class="settings" method="post" action="<?php echo esc_attr( $refresh_transients ); ?>">
 					<?php wp_nonce_field( 'gu_refresh_cache' ); ?>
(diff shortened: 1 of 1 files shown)


src/Git_Updater/Settings.php
```diff
@@ -297,15 +297,15 @@ public function create_admin_page() {
...
+				<a href="https://github.com/afragen/git-updater" target="_blank"><img src="<?php echo esc_url( $logo ); ?>" alt="Git Updater logo" /></a><br>
 				<?php esc_html_e( 'Git Updater', 'git-updater' ); ?>
 				<span class="description"><?php echo esc_html( ' v' . $this->get_plugin_version() ); ?></span>
 			</h1>
 			<?php $this->options_tabs(); ?>
 			<?php $this->admin_page_notices(); ?>
 			<?php if ( 'git_updater_settings' === $tab ) : ?>
 				<?php $this->options_sub_tabs(); ?>
-				<form class="settings" method="post" action="<?php echo esc_attr( $action ); ?>">
+				<form class="settings" method="post" action="<?php echo esc_url( $action ); ?>">
```
afragen: This one isn't really a complete URL it's a fragment which is text.

thisismyurl:
You're right — $action is 'options.php' or 'edit.php?action=git-updater', a controlled relative path rather than a dynamic URL, so esc_attr() is safe there. Happy to revert that one instance. Are the img src and add_query_arg() href changes acceptable as-is?

afragen:
The other changes are good. If you make the update I'll be happy to merge. Thanks.

src/Git_Updater/Settings.php
```diff
@@ -320,7 +320,7 @@ public function create_admin_page() {
 					?>
 				</form>
 				<?php $refresh_transients = add_query_arg( [ 'git_updater_refresh_transients' => true ], $action ); ?>
-				<form class="settings" method="post" action="<?php echo esc_attr( $refresh_transients ); ?>">
+				<form class="settings" method="post" action="<?php echo esc_url( $refresh_transients ); ?>">
```
afragen: Actually this is a string too. The base "URL" is `$action`, which is a string.

afragen (changes requested):

afragen:
The `$refresh_transients` is also just a string as it's based on `$action`. 

So the PR should only have one change.

afragen (approved):

PR #1164, opened 2026-07-13
Create an autofill for URI and branch if an Oauth connection exists

petertwise:
This is totally just an initial proof of concept... but I thought once we have oAuth connection to an account, we can relieve some of the tediousness of having to locate and copy and paste the exact github.com repo URL, remember the exact branch name and if it's master or main for the primary, etc. 

This code is fully Claude generated and should be human reviewed and tested in more detail before merging.

I think this would be a great enhancement and make installing private plugins and themes way faster and more enjoyable!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,17 @@
 #### [unreleased]
+* add GitHub OAuth-powered autocomplete to the Install Plugin/Theme screen — repo URI field shows a live-filtered dropdown of all repositories (personal + organization) accessible via the connected OAuth account, with debounce, spinner, keyboard navigation (↑ ↓ Enter Esc), and ARIA attributes
+* add branch autocomplete to the Repository Branch field — once a connected-account repo URI is entered, typing in the branch field filters available branches fetched from the GitHub API
+* add branch autofill — when a connected-account repo is selected and its default branch is not `master`, the Repository Branch field is populated automatically
+* add field-hiding — when the entered URI belongs to the connected GitHub account (personal or org), the Remote Repository Host selector and GitHub Access Token field are hidden as they are not needed for OAuth-authenticated installs
+* add `Install::ajax_github_repos()` AJAX handler (`gu_github_repos`) — returns paginated, cached repo list for the connected account, searched by query string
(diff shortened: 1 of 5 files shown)


petertwise:
The main features added are:

* add GitHub OAuth-powered autocomplete to the Install Plugin/Theme screen — repo URI field shows a live-filtered dropdown of all repositories (personal + organization) accessible via the connected OAuth account, with debounce, spinner, keyboard navigation (↑ ↓ Enter Esc), and ARIA attributes
* add branch autocomplete to the Repository Branch field — once a connected-account repo URI is entered, typing in the branch field filters available branches fetched from the GitHub API
* add branch autofill — when a connected-account repo is selected and its default branch is not `master`, the Repository Branch field is populated automatically
* add field-hiding — when the entered URI belongs to the connected GitHub account (personal or org), the Remote Repository Host selector and GitHub Access Token field are hidden as they are not needed for OAuth-authenticated installs

petertwise:
Note that the main field is still kept as a text field in both cases, so if autofill doesn't meet your current need at any time for any reason you can still manually enter the value you need. It's not replacing the text field with a <select> where custom values are no longer able to be input.

afragen:
This is intriguing. What happened to the Remote Repository Host and does this also work with the API Add-On plugins?

petertwise:
> What happened to the Remote Repository Host

I hid Remote Repository Host and GitHub Access Token on purpose, on the assumption that if you're using an oAuth connected github repo that's all taken care of already. But I'm not super familiar with any edge case uses for these, so that's easy to back out if you think it's better to keep showing them.

> does this also work with the API Add-On plugins?

I don't think it would work as-is, but would be pretty easy to do something similar for each.
I've only tested it with Github so far as that's what we use.

afragen:
You probably should show both. The Repository Host selector ensures the proper API is used and the token allows for private repositories from other developers.

petertwise:
Yes, they show back up if you use a value that's not one of your own oAuth connected repos, so the support for other developers private repos should still be there...

But sounds like it's easier to just ditch the show/hide completely. I'll make that change and repost here!

afragen:
Consider it a request to make it more universal for any API Add On if installed. I'd rather not have to replicate everything for each Add On. Especially as I don't want multiple JS scripts fighting with each other.

